PR #4, opened 2017-10-04
[SDK-120]: Removed integration tests

echarrod:
Removed integration tests, as they were failing after the changes made which invalidated the token after it's one time use. We will reinstate once we have a sandbox environment to test them on.


echarrod:
Didn't remove the "encrypted_yoti_token.txt", as it's used by other tests which don't contact the Live API

PR #6, opened 2017-10-04
[SDK-39]: Changed to version 2.0.0, after addition of X-SDK headers

echarrod:
Changed to version 2.0.0, after addition of X-SDK headers and added keywords. Similar to the .NET version, I decided to be consistent with the versioning there - even though we've changed the name of the package and could feasibly start at 1.0.0 as it's a new package, there is a smaller chance of confusion if we use 2.0.0. I'll publish the package once this change is approved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,4 @@ ENV/
 .idea/
 *.un~
 .history/
+.vscode
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ This SDK works with Python 2.6+ and Python 3.3+.
 To import the Yoti SDK inside your project, simply run the following command from your terminal:
 
 ```shell
-$ pip install yoti-python-sdk
+$ pip install yoti
 ```
 
 ## SDK Project import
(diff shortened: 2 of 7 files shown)


echarrod:
I'll change the readme in this PR too, so I'll be adding at least another commit to it

echarrod:
I'll also change the install command which is in the integration guide documentation once the package has been published

vassyz:
Having the version number consistent across the SDKs would be ideal, but I think it's quite hard to achieve, especially since we will have bugfix releases for individual issues. I'm not that bothered about it.

echarrod:
It wasn't really the version number consistent across the SDKs that I was talking about - that would be very hard to do! I was talking about the consistency (with .NET) in naming this version 2.0.0, even though technically it could be 1.0.0, as it's a new package.

vassyz:
Removed my commit from master. It has already been pulled into this branch. It looks a bit weird to me, but hopefully, it's all good now. There are also a few references to the old package name in the examples.

echarrod:
I've added requirements changes for the example projects in that last commit. There are instances of import "yoti_python_sdk" in the README and the example projects, but these refer to the internal folder name in the project.

echarrod:
After having a lot of issues with uploading the new package ([SO Issue](https://stackoverflow.com/questions/46606692/cant-upload-to-pypi-with-twine/46644300#46644300) ,
 [SourceForge Issue](https://sourceforge.net/p/pypi/support-requests/768/), I managed to get it working!

PR #7, opened 2017-10-13
[SDK-128]: Renamed X-SDK to X-Yoti-SDK

echarrod:


vassyz:
Can you check if there's a way in Python to get the latest release of a package without mentioning the exact version? I'd want to avoid having to change the example dependencies every time.

echarrod:
I Googled a bit and there doesn't seem to be a way. It kind of makes sense - people sometimes freeze the versions so you can only use certain versions of a package to ensure that they work, but if you just specify the latest, there's no guarantee that it will work. It's only two command line calls to update them, not too hard, although a little annoying.

PR #8, opened 2017-10-16
[SDK-131]: Updated License

echarrod:


vassyz:
You'll also have to change the `setup.py` file to mention MIT.

echarrod:
Good point, I've added another commit now. Includes "OSI Approved" as per the [PyPI specifications](https://pypi.python.org/pypi?%3Aaction=list_classifiers)

PR #12, opened 2017-11-03
[SDK-163]: Tidy up readme

echarrod:
Thought I'd create a PR as wanted to check there's nothing in the old 'Profile Retrieval' section you'd like to keep - it focuses more on the token used to get the profile. In the new version we just have a reference that the profile is retrieved with the token, do you think we need to stress this more? If we were to, would it belong in a different section, so that this 'Handling Users' section is consistent across all the SDKs?


README.md
```diff
@@ -113,21 +113,31 @@ def auth():
...
 ```python
-activity_details = client.get_activity_details(request.args['token'])
-user_profile = activity_details.user_profile
-    return render_template('profile.html',
-                           **user_profile)
-
+client = Client(YOTI_CLIENT_SDK_ID, YOTI_KEY_FILE_PATH)
+activity_details = client.get_activity_details(token)
+
+user_id = activity_details.user_profile.get('user_id')
```
vassyz: I think moving `activity_details.user_profile` into a separate object looks better:

`user_profile = activity_details.user_profile`

echarrod: Yes true, I'll change that

PR #16, opened 2017-12-12
Fix wrong config.py location

jonatasbaldin:


echarrod:
Hi Jonatas, thanks for pointing this out and correcting it. 
If you need any help integrating with Yoti, don't hesitate to contact us on GitHub, or at sdksupport@yoti.com.
Regards,
Ed

PR #18, opened 2018-01-10
[SDK-158]: Added docker files for Flask example application

echarrod:
I'll do the Django project next in the same way

### DIFF
--- a/examples/yoti_example_flask/Dockerfile
+++ b/examples/yoti_example_flask/Dockerfile
@@ -1,0 +1,11 @@
+FROM python:3
+ARG YOTI_APPLICATION_ID
+ARG YOTI_CLIENT_SDK_ID
+ARG YOTI_FULL_KEY_FILE_PATH
+ADD . /code
+WORKDIR /code
+RUN pip install --no-cache-dir -r requirements.txt
+RUN sed -i "s|environ.get('YOTI_APPLICATION_ID')|'$YOTI_APPLICATION_ID' |g" settings.py
+RUN sed -i "s|environ.get('YOTI_CLIENT_SDK_ID')|'$YOTI_CLIENT_SDK_ID' |g" settings.py
+RUN sed -i "s|environ.get('YOTI_KEY_FILE_PATH')|'$YOTI_FULL_KEY_FILE_PATH' |g" settings.py
+CMD ["python", "app.py"]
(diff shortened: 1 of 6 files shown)


examples/yoti_example_flask/app.py
```diff
@@ -44,4 +45,4 @@ def auth():
 
 
 if __name__ == '__main__':
-    app.run(debug=True)
+    app.run(host="0.0.0.0", debug=True)
```
echarrod: This change was applied so that the docker app could be accessed by the host browser. I tested that this still works when not using Docker too.

examples/yoti_example_flask/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3
+ARG YOTI_APPLICATION_ID
+ARG YOTI_CLIENT_SDK_ID
+ARG YOTI_FULL_KEY_FILE_PATH
+ADD . /code
+WORKDIR /code
+RUN pip install --no-cache-dir -r requirements.txt
+RUN sed -i "s|environ.get('YOTI_APPLICATION_ID')|'$YOTI_APPLICATION_ID' |g" settings.py
+RUN sed -i "s|environ.get('YOTI_CLIENT_SDK_ID')|'$YOTI_CLIENT_SDK_ID' |g" settings.py
+RUN sed -i "s|environ.get('YOTI_KEY_FILE_PATH')|'$YOTI_FULL_KEY_FILE_PATH' |g" settings.py
```
sidibos: Maybe you should check if these arguments have been set by the user, for example YOTI_APPLICATION_ID is an integer.

echarrod:
Good point Mousse, I've added checks for each variable now, which will throw an error if they are not set.

I also added a note to the README to instruct the user to put the .pem file within the example folder where docker is being run from.

vassyz:
The issue with putting the Yoti keys in the Docker compose file is that it will make a change tracked by git. The example already uses the `.env` file, meaning you could just take advantage of those environment variables like so:

```
        YOTI_APPLICATION_ID: "${YOTI_APPLICATION_ID}"
        YOTI_CLIENT_SDK_ID: "${YOTI_CLIENT_SDK_ID}"
        YOTI_FULL_KEY_FILE_PATH: "${YOTI_KEY_FILE_PATH}"
```

The only difference is that the example will support any paths to the PEM file, but Docker will only allow files inside the example folder.

echarrod:
Didn't realise you could do that with the docker-compose file, I've changed it now, so it uses the existing .env variables

PR #19, opened 2018-01-17
 [SDK-158]: Added Docker files for Flask, minor tweaks

echarrod:
Flask seems to be slightly different to Django in that the ENV variables need to be explicitly passed in with the DOCKERFILE. Made the changes suggested by Seb in SDK-224 to the README.


README.md
```diff
@@ -129,20 +129,21 @@ nationality = user_profile.get('nationality')
 
 ## Running the Examples
 
-The callback URL for both example projects will be http://localhost:5000/yoti/auth/
+The callback URL for both example projects will be ``http://localhost:5000/yoti/auth/``
```
vassyz: Any particular reason why you would use double ``?

README.md
```diff
@@ -129,20 +129,21 @@ nationality = user_profile.get('nationality')
...
 
 ### With Docker
 
-#### Flask
-
-To run the Flask container:
+To run the Flask or Django containers:
```
vassyz: Should `containers` be plural here?

README.md
echarrod: I misunderstood your comment 
> Please remove the URL and display it as code using ``

I'll change it to use single ones

README.md
echarrod: I think either works. I suppose containers could potentially imply that there were more than one container for each example, so I can change.

examples/yoti_example_django/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3
+ARG YOTI_APPLICATION_ID
+ARG YOTI_CLIENT_SDK_ID
+ARG YOTI_KEY_FILE_PATH
+RUN if [ "$YOTI_APPLICATION_ID" = "" ] ; then echo YOTI_APPLICATION_ID not provided; exit 1; else echo YOTI_APPLICATION_ID is $YOTI_APPLICATION_ID ; fi
+RUN if [ "$YOTI_CLIENT_SDK_ID" = "" ] ; then echo YOTI_CLIENT_SDK_ID not provided; exit 1; else echo YOTI_CLIENT_SDK_ID is $YOTI_CLIENT_SDK_ID ; fi
+RUN if [ "$YOTI_KEY_FILE_PATH" = "" ] ; then echo YOTI_KEY_FILE_PATH not provided; exit 1; else echo YOTI_KEY_FILE_PATH is $YOTI_KEY_FILE_PATH ; fi
```
sidibos: Nice one - but this check would work if the user doesn't change the default arg value e.g YOTI_APPLICATION_ID=yourApplicationId.
Maybe check the value is different from the default value 'yourApplicationId' or change the default value to NULL or empty string.

README.md
```diff
@@ -129,20 +129,21 @@ nationality = user_profile.get('nationality')
...
-1. Change directory to the Flask example project with `cd examples/yoti_example_flask`
+1. Change directory to the example project directory with
+   * `cd examples/yoti_example_flask` for __Flask__   
+   __OR__
+   * `cd examples/yoti_example_django` for __Django__
 1. Make sure the environment variables `YOTI_APPLICATION_ID`, `YOTI_CLIENT_SDK_ID` and `YOTI_KEY_FILE_PATH` are set (instructions in the [Configuration](#configuration) section). _Please note that with Docker, the .pem file must reside in a location within where docker is being run from, so it should be placed somewhere under the respective  [yoti_example_flask](/examples/yoti_example_flask)/[yoti_example_django](/examples/yoti_example_django) folders._
 1. Rebuild the images if you have modified the docker-compose.yml file with
    - `docker-compose build --no-cache`
-1. Build the container with `docker-compose up`
+1. Start the container with `docker-compose up`
```
sidibos: This will work fine, but I kind of like the detached mode `docker-compose up -d` - just a preference.

echarrod: I like to see the logs in the terminal!

vassyz: I prefer it without the `-d` 😏

examples/yoti_example_django/Dockerfile
echarrod: Ah yes, I updated the .env file without updating this check, I'll change to check against `yourApplicationId` and the other similar ones. I'll do the same for the flask Dockerfile too

PR #20, opened 2018-01-19
[SDK-230]: Changed image handling so that the conversion to base64 is done via a helper method

echarrod:


vassyz:
Can you try running the Django example without Docker. On my machine it returns `File "/yoti-python-sdk/examples/yoti_example_django/yoti_example/views.py", line 10, in <module>
    from app_settings import (
ImportError: cannot import name YOTI_KEY_FILE_PATH`.

echarrod:
Works for me, can you send me what your app_settings.py file looks like?

echarrod:
It should have been updated so the name is consistent so now it looks like `YOTI_KEY_FILE_PATH = environ.get('YOTI_KEY_FILE_PATH')`

vassyz:
I pulled the branch, so the `app_settings.py` should be identical to GitHub's. If this PR can wait until I get back, I can double check.

PR #21, opened 2018-02-06
[SDK-260]: Added new profile attributes

echarrod:


README.md
```diff
@@ -120,8 +120,10 @@ user_id = user_profile.get('user_id')
 selfie = user_profile.get('selfie')
 given_names = user_profile.get('given_names')
 family_name = user_profile.get('family_name')
+full_name = user_profile.get('full_name')
 phone_number = user_profile.get('phone_number')
 date_of_birth = user_profile.get('date_of_birth')
+age_verified = user_profile.get('age_verified')
```
vassyz: I don't think this looks alright here. We should remove this line and only rely on the example for age verification. If you copy and paste this code it will not work as `age_verified` is a computed attribute.

echarrod: Ah yes you're right, I'll remove it

PR #22, opened 2018-03-02
[SDK-250]: AML Requests

echarrod:


yoti_python_sdk/client.py
```diff
@@ -69,31 +74,64 @@ def get_activity_details(self, encrypted_request_token):
...
+        aml_profile_json = json.dumps(aml_profile.__dict__)
+        aml_profile_bytes = aml_profile_json.encode()
+        path = self.__endpoint.get_aml_request_url()
+        url = yoti_python_sdk.YOTI_API_ENDPOINT + path
+        headers = self.__get_request_headers(path, http_method, aml_profile_bytes)
 
-        return '/profile/{0}?nonce={1}&timestamp={2}&appId={3}'.format(
-            token, nonce, timestamp, self.sdk_id
-        )
+        response = requests.post(url=url, headers=headers, data=aml_profile_bytes)
```
vassyz: This will become a bit repetitive if we end up having a lot of calls, but it's fine for now.

yoti_python_sdk/tests/test_client.py
```diff
@@ -155,3 +156,35 @@ def test_requesting_activity_details_with_missing_profile(
...
+def test_creating_request_content_is_added(
+        mock_uuid4, mock_time, mock_get, client, expected_headers):
+    content = '{"Content"}'
+    content_bytes = content.encode()
+    request = client._Client__create_request(http_method="GET", path=YOTI_API_ENDPOINT, content=content_bytes)
+
+    b64encoded = base64.b64encode(content_bytes)
+    b64ascii = b64encoded.decode('ascii')
+
+    assert request.endswith("&" + b64ascii)
```
vassyz: Shouldn't there be a test for `perform_aml_check`?

vassyz:
There should also be a mention of running the AML example in here: https://github.com/getyoti/yoti-python-sdk/tree/AML#running-locally

vassyz:
The AML example also needs a `requirements.txt` file with yoti and dotenv. It will obviously not work until you publish the package.

PR #23, opened 2018-03-28
[SDK-282]: Added IsAgeVerified helper

echarrod:


examples/yoti_example_flask/templates/profile.html
```diff
@@ -60,10 +60,10 @@ <h3><a href="/">Home</a></h3>
             </tr>
         {% endif %}
 
-        {% if age_verified %}
+        {% if is_age_verified is defined %}
```
echarrod: Flask used `Jinja`, which uses slightly different syntax to what we're using in the Django project (http://jinja.pocoo.org/docs/2.10/templates/#defined)

yoti_python_sdk/tests/test_activity_details.py
```diff
@@ -0,0 +1,59 @@
...
+
+from yoti_python_sdk.activity_details import ActivityDetails
+from yoti_python_sdk.protobuf.v1.protobuf import Protobuf
+from yoti_python_sdk.tests.conftest import successful_receipt
+
+
+def create_selfie_field(activity_details):
+    activity_details.field = lambda: None
+    activity_details.field.name = "selfie"
+    activity_details.field.value = "base64(ง •̀_•́)ง"
```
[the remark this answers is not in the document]
echarrod: Done

examples/yoti_example_flask/app.py
```diff
@@ -49,4 +46,4 @@ def auth():
 
 
 if __name__ == '__main__':
-    app.run(host="0.0.0.0", debug=True)
+    app.run(host="0.0.0.0")
```
echarrod: debug doesn't work when using Python 2.7. This is just the debug command sent when using the `werkzeug` package, which provides more output from this package, we can still debug fine with Flask. I attempted to switch to a different command here, but this then breaks the Docker process

yoti_python_sdk/tests/conftest.py
```diff
@@ -53,5 +70,5 @@ def x_yoti_auth_digest_get():
 
 @pytest.fixture(scope='module')
 def x_yoti_auth_digest_post():
-    with open(AUTH_DIGEST_POST_FILE_PATH, 'r') as auth_digest_file:
+    with io.open(AUTH_DIGEST_POST_FILE_PATH, mode='r', encoding='utf-8') as auth_digest_file:
```
echarrod: This was the issue we experienced where the test was failing when using Python 2.7: the encoding was defaulting to something else, which meant the digest it was comparing in the test was different. `io.open` is needed to ensure compatibility with 2.7

PR #27, opened 2018-06-20
SDK-378: Self-signed certificates for example projects

echarrod:
- Using pip-require to give clarity on package dependencies
- Updated both example projects to generate on-the-fly self-signed certificates
- Changed dockerfile so it works with HTTPS


README.md
```diff
@@ -209,26 +209,26 @@ To run the Flask or Django container:
...
 
-* Installing dependencies: `pip install -e .[examples]` (If you're using `zsh` you need to escape the square brackets: `pip install -e .\[examples\]`)
-
 #### Flask
 
 1. Change directories to the Flask project: `cd examples/yoti_example_flask`
+1. Install dependencies: `pip install -r requirements.txt`
 1. Run `python app.py`
-1. Navigate to http://localhost:5000
+1. Navigate to https:////localhost:5000
```
sidibos: There are 2 extra slashes here.

echarrod: Good spot, will remove

PR #29, opened 2018-07-30
SDK-442: Anchors

echarrod:
- Updated protobuf classes
- Added anchors
- `test_activity_details.py` and `test_client.py` are not loaded by default, but they still contain important changes, so please make sure to expand them
- I will add the anchors, sources & verifiers to the examples and the documentation in a separate branch and PR, as this was getting quite large


yoti_python_sdk/activity_details.py
```diff
@@ -42,33 +47,48 @@ def try_parse_selfie_field(self, field):
...
+            decoder.decode(field.value),
+            anchors)
+
+    def set_address_to_be_formatted_address_if_null(self, anchors):
+        if config.ATTRIBUTE_POSTAL_ADDRESS not in self.user_profile and config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS in self.user_profile:
+            if config.KEY_FORMATTED_ADDRESS in self.user_profile[config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS]:
+                self.user_profile[config.ATTRIBUTE_POSTAL_ADDRESS] = self.user_profile[config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS][
+                    config.KEY_FORMATTED_ADDRESS]
+
+        if config.ATTRIBUTE_POSTAL_ADDRESS not in self.profile and config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS in self.profile:
```
qzagarese: My understanding is that you are managing backward compatibility with user_profile here and that's why we twice the same condition with two objects. I'm not sure whether it's worth refactoring it, but a comment would help, because it is a bit tough to read.

yoti_python_sdk/anchor.py
```diff
@@ -0,0 +1,108 @@
...
+                    extensions = cert.extensions[i]
+                    if hasattr(extensions, 'oid'):
+                        oid = extensions.oid
+                        if hasattr(oid, 'dotted_string'):
+                            if oid.dotted_string == SOURCE_EXTENSION:
+                                anchor_type = config.ANCHOR_SOURCE
+                            elif oid.dotted_string == VERIFIER_EXTENSION:
+                                anchor_type = config.ANCHOR_VERIFIER
+                            else:
+                                continue
```
qzagarese: Please do not use `continue` as it has tricky semantics. Check whether `anchor_type` is initialised and execute the following code based on that.

yoti_python_sdk/anchor.py
```diff
@@ -0,0 +1,108 @@
...
+
+            if hasattr(anc, 'origin_server_certs'):
+                origin_server_certs_list = list(anc.origin_server_certs)
+                origin_server_certs_item = origin_server_certs_list[0]
+
+                cert = crypto.load_certificate(OpenSSL.crypto.FILETYPE_ASN1, origin_server_certs_item).to_cryptography()
+
+                for i in range(len(cert.extensions)):
+                    extensions = cert.extensions[i]
+                    if hasattr(extensions, 'oid'):
```
qzagarese: this method has a lot of if nesting.
I suggest that you encapsulate the processing of a single extension in a separate method.

yoti_python_sdk/activity_details.py
```diff
@@ -1,16 +1,17 @@
...
+from yoti_python_sdk.anchor import Anchor
+from yoti_python_sdk import config, attribute
 from yoti_python_sdk.protobuf.v1.protobuf import Protobuf
 
 
 class ActivityDetails:
     def __init__(self, receipt, decrypted_profile=None):
         self.decrypted_profile = decrypted_profile
-        self.user_profile = {}
+        self.user_profile = {}  # will be deprecated in v3.0.0
```
sidibos: Do you mean it will be removed from v3.0.0?
If that's the case, the message should be `deprecated`.

yoti_python_sdk/activity_details.py
```diff
@@ -19,19 +20,23 @@ def __init__(self, receipt, decrypted_profile=None):
                     field.value,
                     field.content_type
                 )
-                self.user_profile[field.name] = value
+
+                anchors = Anchor().parse_anchors(field.anchors)
+
+                self.profile[field.name] = attribute.attribute(field.name, value, anchors)
```
sidibos: Is it possible to import `from yoti_python_sdk.attribute import attribute` so we don't have to repeat the name here?

yoti_python_sdk/activity_details.py
```diff
@@ -42,33 +47,48 @@ def try_parse_selfie_field(self, field):
...
-    def try_parse_age_verified_field(self, field):
+    def try_parse_age_verified_field(self, field, anchors):
         if field is not None:
             is_age_verified = Protobuf().value_based_on_content_type(
                 field.value,
                 field.content_type
             )
             if is_age_verified == 'true':
                 self.user_profile['is_age_verified'] = True
+                self.profile['is_age_verified'] = attribute.attribute(is_age_verified, True, anchors)
```
sidibos: Same here `attribute.attribute`

yoti_python_sdk/attribute.py
```diff
@@ -0,0 +1,18 @@
+class attribute:
```
sidibos: Shall we follow the same naming convention as `class Anchor` to start with uppercase for this `class attribute`?

yoti_python_sdk/attribute.py
```diff
@@ -0,0 +1,18 @@
+class attribute:
+    name = ""
+    value = ""
+    anchors = {}
+
+    def __init__(self, name, value, anchors):
```
sidibos: Are we using 'anchors' already instead of 'sources' and 'verifiers'?

yoti_python_sdk/activity_details.py
echarrod: I did think about refactoring, but thought I'd leave it how it would be best once the .user_profile object and associated methods were removed. 

You're right it's tough to read though, I'll add a comment

echarrod:
@sidibos RE: why function names are so long:

Can you think of a more succinct name than `set_address_to_be_formatted_address_if_null `which still describes what the method is doing? I'm of the opinion that a descriptive method name is better than adding comments to the code

yoti_python_sdk/activity_details.py
echarrod: Yes you're right, it should be `removed`, guessing that's what you meant with your second line?

yoti_python_sdk/activity_details.py
echarrod: Yes that makes sense, will change

yoti_python_sdk/activity_details.py
echarrod: Done, updated other instances too

yoti_python_sdk/attribute.py
echarrod: Yes, good spot

yoti_python_sdk/attribute.py
echarrod: We have all 3 available: `.anchors`, `.sources` and `.verifiers` are present, with the latter two just being filters on the former.

yoti_python_sdk/activity_details.py
```diff
@@ -42,33 +48,56 @@ def try_parse_selfie_field(self, field):
...
+        if not isinstance(value_to_decode, str):
+            value_to_decode = value_to_decode.decode()
+
+        self.user_profile[config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS] = decoder.decode(value_to_decode)
+        self.profile[config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS] = Attribute(
+            config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS,
+            decoder.decode(value_to_decode),
+            anchors)
+
+    def set_address_to_be_formatted_address_if_null(self, anchors):
```
bucky-boy: It is certainly better to have descriptive names than comments.  But if the names start to resemble a sentence, it might be going a bit far.  Could this be named something like "set_fallback_address", "ensure_postal_address" or "populate_from_structured_address"?

PR #31, opened 2018-08-15
SDK-492: Profile attributes

echarrod:
- Changed profile to be a class rather than a dict
- Added attribute helpers
- Since Django & Flask templates take dictionaries, we have to cast it to a dict when using these
- Separated the profile parsing logic, so deprecating user_profile will just be a case of removing one of these. There are still tests for both ways currently


yoti_python_sdk/profile.py
```diff
@@ -0,0 +1,120 @@
...
+            if is_age_verified == 'false':
+                self.profile[config.ATTRIBUTE_IS_AGE_VERIFIED] = Attribute(is_age_verified, False, anchors)
+                return
+
+        raise TypeError("age_verified_field unable to be parsed")
+
+    def try_convert_structured_postal_address_to_dict(self, field, anchors):
+        decoder = json.JSONDecoder(object_pairs_hook=collections.OrderedDict, strict=False)
+        value_to_decode = field.value
+        if not isinstance(value_to_decode, str):
```
bucky-boy: Do we need this?  Don't we know the type before we call the method?

examples/yoti_example_flask/templates/profile.html
```diff
@@ -264,21 +264,21 @@ <h3>Yoti Profile</h3>
...
+        {% if profile.gender %}
         <tr>
             <th scope="row">Gender</th>
-            <td>{{ gender.value }}</td>
+            <td>{{ profile.gender.value }}</td>
             <td colspan="3">
                 <table class="table">
                     <tbody>
-                    {% for source in gender.sources.items %}
+                    {% for source in profile.gender.sources.items %}
```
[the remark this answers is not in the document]
echarrod: It has the same effect, but I'll remove it for consistency

yoti_python_sdk/profile.py
```diff
@@ -0,0 +1,120 @@
...
+                    field.value,
+                    field.content_type
+                )
+
+                anchors = Anchor().parse_anchors(field.anchors)
+
+                self.profile[field.name] = Attribute(field.name, value, anchors)
+                # self.profile[field.name] = Attribute(field.name, field.value, anchors)
+
+                if field.name.startswith(config.ATTRIBUTE_AGE_OVER) or field.name.startswith(
```
sidibos: I would rather use a regex here to capture `age_(Over|Under)` as there could be more derivation on age in the future.

yoti_python_sdk/activity_details.py
```diff
@@ -48,53 +46,37 @@ def try_parse_selfie_field(self, field):
...
         )
 
-    def try_parse_age_verified_field(self, field, anchors):
+    def try_parse_age_verified_field(self, field):
         if field is not None:
             is_age_verified = Protobuf().value_based_on_content_type(
                 field.value,
                 field.content_type
             )
             if is_age_verified == 'true':
```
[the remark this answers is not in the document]
echarrod: This currently catches the (very unlikely) case that `is_age_verified` is a value other than `true` or `false` - as we would want to throw an error in that case.  Although removing it would make it more concise...

yoti_python_sdk/profile.py
echarrod: Looks like we needed it in a previous iteration of the code, but not anymore. The tests need it at the moment, but the much better approach would be to encode the tests in bytes, and not this way.

yoti_python_sdk/profile.py
echarrod: As discussed, we'll leave this as is, as any future derivations on age are likely to be in a separate attribute, so both of these approaches are fine.

yoti_python_sdk/activity_details.py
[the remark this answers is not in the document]
echarrod: @bucky-boy I think it would be be better to leave it, to be safe

PR #32, opened 2018-08-16
SDK-508: Change is_age_verified to be age_verified

echarrod:
- Also replicated functionality for old profile processing and new, so that we don't throw the error when the attribute fails to parse


README.md
```diff
@@ -127,7 +127,7 @@ family_name = profile.family_name.value
 full_name = profile.full_name.value
 phone_number = profile.phone_number.value
 date_of_birth = profile.date_of_birth.value
-is_age_verified = profile.is_age_verified.value
+age_verified = profile.age_verified.value
```
qzagarese: Is the current version exposing `is_age_verified` ? If so, this is a breaking change and it cannot go out like this.
You need to deprecate `is_age_verified` instead and add `age_verified`.

echarrod: Was just committing as you commented: I have kept the `is_age_verified` for activity_details, and we now use `age_verified` for the profile object. I've tested this on the old version of the Django example project too to check it works. Do you think we should add anything to the API coverage? Or it would just be sufficient to include this information in the release notes?

bucky-boy:
So, @echarrod , following Sprint Planning you're going to change this PR to remove age_verified from the new Profile altogether. Correct?

echarrod:
@bucky-boy correct, will update this PR later today

echarrod:
Closing this one in favour of creating a new PR, as a lot of the changes are reverting what was previously committed

PR #34, opened 2018-11-14
SDK-556 SDK-559 SDK-592: receipt_id, email_address, cryptography

echarrod:
- Although the vulnerability doesn't seem to be showing up anymore on https://github.com/getyoti/yoti-python-sdk/network/dependencies, might as well bump the version of `cryptography` anyway
- Add to activity_details:
  - receipt_id
  - timestamp
- Add email_address getter to new profile class


requirements.txt
```diff
@@ -10,7 +10,7 @@ attrs==18.1.0             # via pytest
 cffi==1.11.5              # via cryptography
 click==6.6
 colorama==0.3.9           # via pytest
-cryptography==2.2.2
+cryptography==2.4.1
```
bucky-boy: Here we have 2.4.1, but in _requirements.in_ it's 2.3.0?

echarrod: requirements.in specifies a _minimum_ version (`cryptography>=2.3.0`), so then when we run `pip-compile`, it just chooses the version of each package which satisfies all of the requirements.

echarrod: Do you think it would be better to target ==2.4.1 explicitly in the requirements.in?

bucky-boy: I don't know this tech in detail.
If _pip-compile_ is trying to find the 'best fit', and if its consistent + reliable, maybe it's best not to interfere with it.
I'm happy to leave this to your expertise...

echarrod: Not really anything in the way of release notes on their [GitHub page](https://github.com/pyca/cryptography/releases)

yoti_python_sdk/tests/test_activity_details.py
```diff
@@ -119,8 +136,9 @@ def test_try_parse_age_verified_field_age_under():
 def test_try_parse_age_verified_field_non_bool_value_not_parsed():
     activity_details = ActivityDetails(successful_receipt())
     create_age_verified_field(activity_details, True, "18".encode(), 18)
-
+    sys.stdout = open(os.devnull, 'w')  # disable print
```
qzagarese: What is the reason for disabling and re-enabling stdout?

echarrod: It just logs the error about parsing the age verified value when it's a non-bool value, so this is just disabling that for the line where we expect it to happen.

requirements.txt
bucky-boy: Please go with this as it is, get the release done.

This is a file generated by our build process.  It's the process we've had in place from the beginning.  If we want to investigate + review it we can do so when we come to the major review (raise a ticket)

PR #40, opened 2019-02-06
SDK-668: SDK Version in request headers

echarrod:
Would be clearer to look at if you look at the commits separately:

- 8714d29
  - Extracting version to own file, which can then be accessed in the tests and when sending the header
  - Extracting headers to constants

- 19890f1
  - Code cleanup using PyCharm built-in functions


yoti_python_sdk/client.py
```diff
@@ -15,7 +15,7 @@
 from yoti_python_sdk.crypto import Crypto
 from yoti_python_sdk.endpoint import Endpoint
 from yoti_python_sdk.protobuf.v1 import protobuf
-from .config import SDK_IDENTIFIER
+from .config import *
```
bucky-boy: Is this is a good idea?  Is it intentional?

yoti_python_sdk/profile.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, profile_attributes):
                 if field.name == config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS:
                     self.try_convert_structured_postal_address_to_dict(field, anchors)
 
-            self.ensure_postal_address(anchors)
+            self.ensure_postal_address()
```
bucky-boy: Why is this changing?

yoti_python_sdk/version.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+__version__ = "2.6.0"
```
bucky-boy: What's the rationale behind putting this in its own file?  Is it somehow part of the build?

yoti_python_sdk/__init__.py
```diff
@@ -9,11 +10,18 @@
     'YOTI_API_VERSION': 'v1',
 }
 
+main_ns = {}
+ver_path = convert_path('yoti_python_sdk/version.py')
```
bucky-boy: So, do we read this file only once, at startup?

yoti_python_sdk/profile.py
echarrod: PyCharm alerted me that it's possible anchors could not be initialised before it gets here. I've changed the behaviour of `ensure_postal_address` to use the structured_postal_address anchors directly there instead

yoti_python_sdk/__init__.py
echarrod: We use it again when sending the headers: `sdk_version = yoti_python_sdk.__version__`, but it's not possible to access the version that way in this class, as it hasn't been initalised yet

yoti_python_sdk/client.py
echarrod: Since I extracted all of the constants out, we would have to do `from .config import X_YOTI_SDK, SDK_IDENTIFIER, X_YOTI_SDK_VERSION, JSON_CONTENT_TYPE, X_YOTI_AUTH_KEY, X_YOTI_AUTH_DIGEST` instead. I thought it is neater this way. 

I know we should be wary of importing all, but given that these are just constants I don't think it's too problematic.

yoti_python_sdk/version.py
echarrod: Yes it is part of the build. If we are wanting to define the version in one place, we need a way of accessing it during `__init__.py`. So during `__init__.py`, we read it from the file, since we need to know the version **before** we initalise it (and then could use `yoti_python_sdk.__version__`. 

I didn't think it would be efficient or necessary to put anything else in this file.

yoti_python_sdk/__init__.py
bucky-boy: Can this be refactored so that we only read from disk once, when the client is initialised?  
More File IO each time we request a profile really isn't a good thing.

echarrod: `__init.py__` is only run once, to initialize the package ([link to docs](https://docs.python.org/3/tutorial/modules.html#packages)). From then on, it's part of a variable which is available package-wide, but we only read from disk once

davidgrayston:
@echarrod I ran the Flask example and debugged `headers` variable locally - I can see `'X-Yoti-SDK-Version': '2.6.0'` along with existing headers (and the share was successful)

I had an issue running the example as the version file is relative to the SDK code, which meant I had to run from the root of the project like this:
```
$ python examples/yoti_example_flask/app.py
```

In order to run the checked out code to test, I replaced `yoti==2.5.0` with `../../` in the example `requirements.txt`

echarrod:
Thanks @davidgrayston, I've updated the version filepath calculation which should allow it to work from anywhere, so it shouldn't matter where you run it from

davidgrayston:
I'm now able to run this from the `yoti_example_flask` directory `python app.py` and seeing header `'X-Yoti-SDK-Version': '2.6.0'`

PR #41, opened 2019-02-11
SDK-778: Update Django required version 

echarrod:
- Update Django required version to prevent compatibility issue with Python 3.7+


echarrod:
@bucky-boy this is just updating the required Django version for the example project. The change to `setup.py` is in the `extras_require` section, which means it will not be installed by default unless needed

PR #46, opened 2019-03-26
SDK-781: Add timezone information

echarrod:
See here for information on Python DateTime - https://docs.python.org/3/library/datetime.html

Our time zones were "naive", which meant they had no timezone information attached to them. This meant the tests were failing when run in a different timezone. This change adds the fact that they are UTC to the timezones, so they are now "aware".

We have to think about whether anyone may have been making any assumptions about the timezone, and what this change would mean if the were.

### DIFF
--- a/yoti_python_sdk/activity_details.py
+++ b/yoti_python_sdk/activity_details.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import collections
 import json
-from datetime import datetime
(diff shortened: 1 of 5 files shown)


echarrod:
On second thoughts, this will be a breaking change. However small the possibility of it affecting clients, it's best to wait until the major release to do this. Will close for now, and fix the tests another way

PR #48, opened 2019-03-26
SDK-879: Travis

echarrod:
- OSX and Windows not currently supported :|
- 3.8-Dev seems currently unreliable, so allowing failures for it
- Had a very tricky issue with older versions of Python ordering the JSON differently, and so the X-Yoti-Digest calculated was different, as it was a different payload - solved this by sorting the keys before adding to the payload. 
- Removing the datetime comparison until we have implemented SDK-891


yoti_python_sdk/config.py
```diff
@@ -23,3 +25,5 @@
 X_YOTI_SDK = "X-Yoti-SDK"
 X_YOTI_SDK_VERSION = X_YOTI_SDK + "-Version"
 JSON_CONTENT_TYPE = 'application/json'
+UTC_TIMEZONE = pytz.utc
```
davidgrayston: Will this only affect the tests?

echarrod: Yes, good point, it shouldn't be in the SDK if it's only used in tests. It's not actually used after the latest changes, so I'll remove it altogether.

PR #49, opened 2019-03-28
SDK-868: Add INT support, log when error parsing attribute, formatter

echarrod:
- added INT support
- adding logging when there was an error parsing the attribute
- added tests around expected format
- made it clearer what we should be doing with PNG and JPEG types


echarrod:
.travis.yml file mentioned in README isn't present in this branch, will be available once https://github.com/getyoti/yoti-python-sdk/pull/48 is merged

yoti_python_sdk/profile.py
```diff
@@ -14,10 +15,15 @@ def __init__(self, profile_attributes):
...
-                value = Protobuf().value_based_on_content_type(
-                    field.value,
-                    field.content_type
-                )
+                try:
+                    value = Protobuf().value_based_on_content_type(
+                        field.value,
+                        field.content_type
+                    )
+                except Exception as exc:
```
davidgrayston: This will catch `raise TypeError('Wrong content type')` which means attributes with the `UNDEFINED` content type will be returned as an Attribute object with a `None` value - is this a breaking change?

We now return attributes with `UNDEFINED` content type as `NULL` in the PHP SDK - should this change in behaviour remain as it is until we standardise the handling on the `UNDEFINED` content type.

Perhaps creation of `Attribute(field.name, value, anchors)` should be moved into the try block? Or alternatively, we change the behaviour in the PHP SDK.

echarrod: > This will catch `raise TypeError('Wrong content type')` which means attributes with the `UNDEFINED` content type will be returned as an Attribute object with a `None` value - is this a breaking change?

I think we have to treat this in a similar way to PHP - assuming we never had a protobuf type of undefined sent through from Core, I wouldn't count it as a breaking change if we are changing the behaviour of undefined content types. 

> We now return attributes with `UNDEFINED` content type as `NULL` in the PHP SDK - should this change in behaviour remain as it is until we standardise the handling of the `UNDEFINED` content type?

We can discuss this on Monday, I won't merge this PR until we have agreed on a common approach. 

> Perhaps creation of `Attribute(field.name, value, anchors)` should be moved into the try block? Or alternatively, we change the behaviour in the PHP SDK.

I did have it as that originally, but I thought I didn't want failing to parse the attribute affecting whether the anchors were parsed. On second thoughts, I think I should add a different try/catch block to parsing the anchors - then one of anchors/values failing won't affect the parsing of the other one. What do you think?

setup.py
```diff
@@ -30,14 +30,12 @@
         'Operating System :: OS Independent',
         'Intended Audience :: Developers',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.6',
-        'Programming Language :: Python :: 2.7',
```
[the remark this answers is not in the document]
echarrod: I was mistaken in adding this - I thought we didn't support it - but I remember now we just had a **dev** dependency that means you can't use 2.7, not from a consumer perspective. Clients can still consume the SDK with 2.7. Have added it to the Travis pipeline and tested locally with 2.7 successfully, so will remove this.

yoti_python_sdk/profile.py
davidgrayston: Something like this?

```python
try:
    value = Protobuf().value_based_on_content_type(
      field.value,
      field.content_type
    )
    anchors = Anchor().parse_anchors(field.anchors)
    attr = Attribute(field.name, value, anchors)
except Exception as exc:
    error = 'Error parsing profile attribute: "{0}"'.format(field.name)
    logging.warning('error: {0}, exception: {1} - {2}'.format(error, type(exc).__name__, exc))
    attr = None

self.attributes[field.name] = attr
```

This should then return `None` for attributes that have `UNDEFINED` content type

README.md
```diff
@@ -277,7 +277,7 @@ Running the tests is done by the following process, ensuring you are using Pytho
 1. Install the SDK: `python setup.py develop`
 1. Execute in the main project dir: `py.test`
 
-For information on testing with multiple Python versions, see [VERSION-SUPPORT.md](/VERSION-SUPPORT.md)
```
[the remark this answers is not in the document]
echarrod: Yes I just put a comment on the PR itself: `.travis.yml file mentioned in README isn't present in this branch, will be available once #48 is merged`. Would have probably have been better to have included this in that PR, but I don't think it's worth reverting this commit here.

yoti_python_sdk/profile.py
davidgrayston: Or separate try/catch like this?

```python
try:
    anchors = Anchor().parse_anchors(field.anchors)
except Exception as exc:
    error = 'Error parsing anchors: "{0}"'.format(field.name)
    logging.warning('error: {0}, exception: {1} - {2}'.format(error, type(exc).__name__, exc))
    anchors = None

try:
    value = Protobuf().value_based_on_content_type(
      field.value,
      field.content_type
    )
    attr = Attribute(field.name, value, anchors)
except Exception as exc:
    error = 'Error parsing profile attribute: "{0}"'.format(field.name)
    logging.warning('error: {0}, exception: {1} - {2}'.format(error, type(exc).__name__, exc))
    attr = None

self.attributes[field.name] = attr
```

echarrod: Have changed it to be like your first suggestion, and not returning an attribute, in fitting what Java does.

yoti_python_sdk/protobuf/protobuf.py
```diff
@@ -40,13 +43,20 @@ def anchor(data):
...
-        if content_type == self.CT_UNDEFINED:
-            raise TypeError('Wrong content type')
-        elif content_type == self.CT_STRING:
+        if content_type == self.CT_STRING:
             return value.decode('utf-8')
         elif content_type == self.CT_DATE:
             return value.decode('utf-8')
-        return value
+        elif content_type == self.CT_JPEG \
+                or content_type == self.CT_PNG:
```
davidgrayston: Could this be `elif content_type in [self.CT_JPEG, self.CT_PNG]`?

yoti_python_sdk/protobuf/protobuf.py
```diff
@@ -40,13 +43,20 @@ def anchor(data):
...
+            return value
+        elif content_type == self.CT_INT:
+            string_value = value.decode('utf-8')
+            int_value = int(string_value)
+            return int_value
```
davidgrayston: Could this be `return int(value.decode('utf-8'))`? _(probably clearer as you have it)_

echarrod: I do think it gives more clarity if we separate it

yoti_python_sdk/protobuf/protobuf.py
echarrod: I'm easy to be honest, either way is fine I think

davidgrayston: I think it's fine as is

PR #52, opened 2019-04-02
SDK-900: Parse all JSON attributes, disable logging in specific tests

echarrod:


echarrod:
As long as we only show warnings when we're not expecting them when running the tests, I'm happy!

We will try to implement something like PHP in the future (https://github.com/getyoti/yoti-php-sdk/pull/58), where we are creating a specific "expected logs" file, which can then be checked, but won't do that in this PR.

PR #55, opened 2019-04-04
SDK-594: new Image type

echarrod:
- Creating the new Image object before we add the DocumentImages attribute
- I have left Image.base64_content calling Protobuf().image_uri_based_on_content_type in this PR, as that has moved on my other branch. When I rebase the other branch I will change the location of this (so will leave the TODO in there until then) 
- Since selfie currently is just an attribute with the byte value as the value (retrieved with selfie.value), and base64URL is retrieved from the ActivityDetails, we can't make a non-breaking change to the object that is returned. When we do the major release, I will change the selfie to return this new Image object.

### DIFF
--- a/yoti_python_sdk/image.py
+++ b/yoti_python_sdk/image.py
@@ -3,11 +3,15 @@ from yoti_python_sdk.protobuf.protobuf import Protobuf
 
 class Image:
     def __init__(self, image_bytes, image_content_type):
-        if image_content_type == Protobuf.CT_JPEG or image_content_type == Protobuf.CT_PNG:
+        if image_content_type in Image.allowed_types():
             self.__data = image_bytes
             self.__content_type = image_content_type
         else:
             raise TypeError("Content type '{0}' is not a supported image type".format(image_content_type))
+
+    @staticmethod
+    def allowed_types():
+        return [Protobuf.CT_PNG, Protobuf.CT_JPEG]
 
     @property
     def data(self):
--- a/yoti_python_sdk/image.py
+++ b/yoti_python_sdk/image.py
@@ -1,0 +1,32 @@
+from yoti_python_sdk.protobuf.protobuf import Protobuf
+
+
+class Image:
+    def __init__(self, image_bytes, image_content_type):
+        if image_content_type == Protobuf.CT_JPEG or image_content_type == Protobuf.CT_PNG:
+            self.__data = image_bytes
+            self.__content_type = image_content_type
+        else:
+            raise TypeError("Content type '{0}' is not a supported image type".format(image_content_type))
+
+    @property
+    def data(self):
+        return self.__data
+
+    @property
+    def content_type(self):
+        return self.__content_type
+
+    def mime_type(self):
+        if self.__content_type == Protobuf.CT_JPEG:
+            return "image/jpeg"
+        elif self.__content_type == Protobuf.CT_PNG:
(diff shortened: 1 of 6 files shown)


yoti_python_sdk/image.py
```diff
@@ -0,0 +1,32 @@
+from yoti_python_sdk.protobuf.protobuf import Protobuf
+
+
+class Image:
+    def __init__(self, image_bytes, image_content_type):
+        if image_content_type == Protobuf.CT_JPEG or image_content_type == Protobuf.CT_PNG:
```
davidgrayston: Could we use a dictionary of allowed image types and their mime type to reduce `if/elif` and `or`? e.g.
```python
self.__image_types = {
    Protobuf.CT_JPEG: 'image/jpeg',
    Protobuf.CT_PNG: 'image/png',
}
```

Then you could do something like:
```python
    def __init__(self, image_bytes, image_content_type):
         if image_content_type in self.__image_types:
         ...
```

```python
    def mime_type(self):
        if self.__content_type in self.__image_types:
            return self.__image_types[self.__content_type]
        else:
            ...
```

yoti_python_sdk/profile.py
```diff
@@ -19,6 +20,12 @@ def __init__(self, profile_attributes):
                         field.content_type
                     )
 
+                    # first condition will be removed in v3.0.0, so selfie also returns an Image object
+                    if field.name != config.ATTRIBUTE_SELFIE and \
+                            (field.content_type == Protobuf.CT_JPEG \
+                             or field.content_type == Protobuf.CT_PNG):
+                        value = Image(field.value, field.content_type)
```
davidgrayston: It probably doesn't make much difference now, but could the conversion to Image object happen in `Protobuf().value_based_on_content_type`, then have this use the image object to get the image content, e.g.

```python
if field.name == config.ATTRIBUTE_SELFIE and isinstance(value, Image):
    value = value.content
```

Then we just remove this check completely in the major release? _(I think it's fine as is, but thought I'd mention for consideration)_

You might need to move Image object conversion into `Protobuf().value_based_on_content_type` for document images

yoti_python_sdk/image.py
echarrod: I like the idea of extracting the "allowed_types", so I will do that. Not sure if it's worth adding the dictionary to include the mime type, as we only use that in one method currently.

yoti_python_sdk/profile.py
echarrod: That would mean I'd have to do it in two separate places - as we have one for the old profile, and one for the new one. But I still think it's a good idea to ensure we don't miss this in the major. Will implement

yoti_python_sdk/tests/test_protobuf.py
```diff
@@ -5,32 +5,27 @@
...
-    result = proto.value_based_on_content_type(byte_value, proto.CT_JPEG)
-    assert result == byte_value
-
-    result = proto.value_based_on_content_type(byte_value, proto.CT_PNG)
-    assert result == byte_value
-
-    result = proto.value_based_on_content_type(byte_value, proto.CT_INT)
-    assert result == int_value
 
+def test_warning_protobuf_values_based_on_content_type(proto):
```
davidgrayston: Should this be something like `test_protobuf_values_based_on_unknown_content_type`?

PR #56, opened 2019-04-09
SDK-765: Document Images attribute

echarrod:
- Add `profile.DocumentImages`
- Extracted functionality out of `Protobuf` into `attribute_parser`
- DocumentImages takes advantage of new Image type
- `Selfie` still returns a byte value for backwards compatibility, this will switch over to an `Image` in the major
- We can easily get the type of the MultiValue in Python, and it can be stored in a mixed type list, so there is no need for a wrapper object.


yoti_python_sdk/multivalue.py
```diff
@@ -0,0 +1,24 @@
...
+    from yoti_python_sdk import attribute_parser  # needed here (and not above) for Python 2.7 & 3.4 dependency handling
+
+    proto = protobuf.Protobuf()
+    multi_value_list = []
+    parsed_multi_value = proto.multi_value(multi_value_bytes)
+
+    for multi_value_item in parsed_multi_value.values:
+        value = attribute_parser.value_based_on_content_type(multi_value_item.data, multi_value_item.content_type)
+
+        if value.content_type == proto.CT_MULTI_VALUE:
```
davidgrayston: Is this check needed? It looks like `multivalue.parse()` would be have already been called in `value_based_on_content_type()`:
```python
    elif content_type == Protobuf.CT_MULTI_VALUE:
        return multivalue.parse(value)
```

yoti_python_sdk/multivalue.py
```diff
@@ -0,0 +1,24 @@
...
+            multi_value_list.append(parse(value))
+        else:
+            multi_value_list.append(value)
+
+    return multi_value_list
+
+
+def filter_values(values, type_to_filter):
+    return list(filter(lambda v: isinstance(v, type_to_filter), values))
```
davidgrayston: Should this be immutable?

echarrod: Yep!

PR #58, opened 2019-04-15
master -> development release v2.7.0

echarrod:
Proposed release notes:

### Added
- `document_images` attribute, accessed through `profile.document_images`, using new `Image` type
- `int` content type (accessed through `profile.get_attribute(attribute_name)`)

### Changed
- SDK Version header value
- regenerated protobuf files, removed version folder

### Removed 
- tox package (in favour of testing with Travis)


echarrod:
Unsure if we should mention 
> Changed:
> - regenerated protobuf files, removed version folder

in the release notes. I thought it might be useful since it is a change to the internal folder structure of the SDK.

PR #67, opened 2019-07-17
SDK-495: new demo design for flask

emmas-yoti:


examples/yoti_example_flask/.env.example
```diff
@@ -1,3 +1,3 @@
-YOTI_APPLICATION_ID=yourApplicationId
+YOTI_APPLICATION_ID=yourScenarioId
```
echarrod:
```suggestion
YOTI_SCENARIO_ID=yourScenarioId
```

examples/yoti_example_flask/templates/profile.html
```diff
@@ -1,382 +1,124 @@
+{% macro parse_documents(prop) %}
```
echarrod: Since we also have a `DocumentDetails` attribute (yet to be added to Python), it might be slightly clearer if this was called either `parse_images` or `parse_document_images` 👍 
```suggestion
{% macro parse_document_images(prop) %}
```

echarrod: Probably best to not commit it here in the GUI, as it'll obviously need to be updated where it's called as well

PR #71, opened 2019-07-22
SDK-722: dynamic scenarios

emmas-yoti:
### Added
 - module dynamic_sharing_service with children policy and extension
 - class dynamic_sharing_service.DynamicScenarioBuilder
 - classes DynamicPolicyBuilder and WantedAttributeBuilder in dynamic_sharing_service.policy
 - classes ExtensionBuilder, LocationConstraintExtensionBuilder and TransactionalFlowExtensionBuilder in dynamic_sharing_service.extension
 - module dynamic_sharing_service.share_url with method create_share_url
 - method make_request added to Client
 - Add /dynamic_share page to both Django and Flask example projects which demonstrates this feature

### Changed
 - Autoformatting client.py using black
 - private Endpoints object in Client given public getter, endpoints, to be used with make_request

### Other notes
 - Unlike Node implementation this PR does not add any objects representing DynamicScenarios/Policy etc... The builders construct key-value trees that can be written directly to the endpoint using make_request and should not be manipulated by users.


yoti_python_sdk/tests/dynamic_sharing_service/policy/test_dynamic_policy_builder.py
```diff
@@ -0,0 +1,119 @@
...
+    policy = (
+        DynamicPolicyBuilder()
+        .with_selfie_auth(True)
+        .with_selfie_auth(False)
+        .with_pin_auth()
+        .with_pin_auth()
+        .build()
+    )
+
+    assert len(policy["wanted_auth_types"]) == 1
```
davidgrayston: @echarrod we don't have coverage for this situation in the Node SDK (I think I raised this whilst copying Java) - the Node SDK will not allow you to add then remove the same auth when building a policy - is this something to standardise one way or another? I believe [Java is the same](https://github.com/getyoti/yoti-java-sdk/blob/master/yoti-sdk-impl/src/main/java/com/yoti/api/client/shareurl/policy/SimpleDynamicPolicyBuilder.java#L186)

davidgrayston: The difference in behaviour is here, where the auth type is always added, then filtered out later:
```python
    def with_wanted_auth_type(self, wanted_auth_type, wanted=True):
        self.__wanted_auth_types[wanted_auth_type] = wanted
        return self
```

(I think I prefer this approach, but do we update other SDKs to follow suit?)

yoti_python_sdk/dynamic_sharing_service/share_url.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def create_share_url(yoti_client, dynamic_scenario):
+    http_method = "POST"
+    payload = json.dumps(dynamic_scenario, sort_keys=True).encode()
+    endpoint = yoti_client.endpoints.get_dynamic_share_request_url()
+    response = yoti_client.make_request(http_method, endpoint, payload)
+
+    status_code = response.status_code
+    if status_code >= 200 and status_code < 300:
```
echarrod: Is this a bit clearer and more concise? 
```suggestion
    if 200 <= status_code < 300:
```

yoti_python_sdk/dynamic_sharing_service/share_url.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import json
+
+INVALID_DATA = "Json is incorrect, contains invalid data"
+APPLICATION_NOT_FOUND = "Application was not found"
+UNKNOWN_ERROR = "Unknown HTTP error occured"
```
echarrod:
```suggestion
UNKNOWN_ERROR = "Unknown HTTP error occurred"
```

yoti_python_sdk/dynamic_sharing_service/share_url.py
```diff
@@ -0,0 +1,46 @@
...
+    status_code = response.status_code
+    if status_code >= 200 and status_code < 300:
+        response_json = json.loads(response.text)
+
+        return ShareUrl(
+            _ShareUrl__qr_code=response_json["qrcode"],
+            _ShareUrl__ref_id=response_json["ref_id"],
+        )
+
+    elif status_code == 400:
```
echarrod: Since this logic would also apply to the other requests we make (AML & Profile), could we extract this logic out into a separate class and utilise it both here and in the other requests? (might have to change the 400 error to be more generic)

yoti_python_sdk/tests/dynamic_sharing_service/policy/test_dynamic_policy_builder.py
```diff
@@ -0,0 +1,119 @@
...
+def test_remember_me():
+    policy = DynamicPolicyBuilder().with_wanted_remember_me().build()
+
+    assert policy["wanted_remember_me"]
+
+
+def test_build_with_simple_attributes():
+    builder = DynamicPolicyBuilder()
+    builder.with_family_name()
+    builder.with_given_names()
```
echarrod: Do you think it looks better with the backslash on the preceding line rather than `builder`?
```suggestion
    .with_given_names()\
```

yoti_python_sdk/tests/dynamic_sharing_service/policy/test_dynamic_policy_builder.py
echarrod: Good spot, have created SDK-1093 to address this

yoti_python_sdk/tests/dynamic_sharing_service/policy/test_wanted_attribute_builder.py
```diff
@@ -0,0 +1,15 @@
...
+)
+
+
+def test_build():
+    NAME = "Test name"
+    DERIVATION = "Test derivation"
+
+    builder = WantedAttributeBuilder()
+
+    attribute = builder.withName(NAME).withDerivation(DERIVATION).build()
```
echarrod:
```suggestion
    attribute = builder
    .withName(NAME)
    .withDerivation(DERIVATION)
    .build()
```

yoti_python_sdk/tests/dynamic_sharing_service/test_dynamic_scenario_builder.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_build_scenario():
+    EXTENSION1 = "Extension 1"
+    EXTENSION2 = "Extension 2"
+    CALLBACK_ENDPOINT = "Callback Endpoint"
+
+    scenario = (
+        DynamicScenarioBuilder()
+        .with_policy(
+            DynamicPolicyBuilder().with_full_name().with_wanted_remember_me().build()
```
echarrod:
```suggestion
            DynamicPolicyBuilder()
            .with_full_name()
            .with_wanted_remember_me()
            .build()
```

yoti_python_sdk/tests/dynamic_sharing_service/test_share_url.py
```diff
@@ -0,0 +1,67 @@
...
+
+@mock.patch("requests.request", side_effect=mocked_requests_post_share_url_invalid_json)
+@mock.patch("time.time", side_effect=mocked_timestamp)
+@mock.patch("uuid.uuid4", side_effect=mocked_uuid4)
+def test_create_share_url_invalid_json(mock_uuid4, mock_time, mock_get):
+    yoti_client = Client(YOTI_CLIENT_SDK_ID, PEM_FILE_PATH)
+    dynamic_scenario = DynamicScenarioBuilder().build()
+
+    try:
+        share_url.create_share_url(yoti_client, dynamic_scenario)
```
echarrod: You can do this instead:
```python
    with pytest.raises(RuntimeError) as exc:
        share_url.create_share_url(yoti_client, dynamic_scenario)
    assert str(exc) == share_url.INVALID_DATA
```

yoti_python_sdk/tests/dynamic_sharing_service/test_share_url.py
```diff
@@ -0,0 +1,67 @@
...
+@mock.patch("time.time", side_effect=mocked_timestamp)
+@mock.patch("uuid.uuid4", side_effect=mocked_uuid4)
+def test_create_share_url_app_not_found(mock_uuid4, mock_time, mock_get):
+    yoti_client = Client(YOTI_CLIENT_SDK_ID, PEM_FILE_PATH)
+    dynamic_scenario = DynamicScenarioBuilder().build()
+
+    try:
+        share_url.create_share_url(yoti_client, dynamic_scenario)
+        assert False  # Should have thrown
+    except RuntimeError as err:
```
echarrod: Same here

examples/yoti_example_django/yoti_example/views.py
```diff
@@ -1,51 +1,80 @@
...
+
+load_dotenv(find_dotenv())
+
 from app_settings import (
     YOTI_SCENARIO_ID,
     YOTI_CLIENT_SDK_ID,
-    YOTI_KEY_FILE_PATH
-)
+    YOTI_KEY_FILE_PATH,
+)  # noqa
```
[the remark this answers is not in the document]
emmas-yoti: loaddotenv being before it/it being a module level import that's not at the start of the file. Because of how app_settings is written, it sets the values of these constants to values loaded in by dotenv.

examples/yoti_example_django/yoti_example/views.py
```diff
@@ -1,51 +1,80 @@
...
-        return self.render_to_response({'scenario_id': YOTI_SCENARIO_ID})
+        return self.render_to_response({"scenario_id": YOTI_SCENARIO_ID})
+
+
+class DynamicShareView(TemplateView):
+    template_name = "dynamic_share.html"
+
+    def get(self, request, *args, **kwargs):
+        client = Client(YOTI_CLIENT_SDK_ID, YOTI_KEY_FILE_PATH)
+        policy = DynamicPolicyBuilder().with_full_name().with_age_over(18).build()
```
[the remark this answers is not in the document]
emmas-yoti: Black splits/joins it automatically based on line length.

yoti_python_sdk/dynamic_sharing_service/share_url.py
emmas-yoti: Oo. I didn't know you could do that.

echarrod: [PyCharm](https://www.jetbrains.com/pycharm/download) (community edition is free) suggests things like that automatically, if you ever want to use anything but Vim 😃

examples/yoti_example_django/yoti_example/views.py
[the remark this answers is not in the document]
echarrod: Fine with leaving this as is, since i've become a `black` convert

echarrod:
Have marked my comments relating to `black` as resolved, since we're going to take the formatting as it is! 👍

yoti_python_sdk/dynamic_sharing_service/policy/dynamic_policy_builder.py
```diff
@@ -0,0 +1,122 @@
...
+        attribute = (
+            WantedAttributeBuilder()
+            .withName(config.ATTRIBUTE_DATE_OF_BIRTH)
+            .withDerivation(derivation)
+            .build()
+        )
+        return self.with_wanted_attribute(attribute)
+
+    def with_age_over(self, age):
+        return self.with_age_derived_attribute(config.ATTRIBUTE_AGE_OVER + str(age))
```
echarrod: Since Python is dynamically typed, could we also put a check here (and in the method below) that the age is actually an integer? Like David has done in Node - https://github.com/getyoti/yoti-node-sdk/blob/development/src/dynamic_sharing_service/policy/dynamic.policy.builder.js#L68. 

Core will prevent this, but we might as well also add a check here, since I can't envisage this ever changing

emmas-yoti: Age doesn't have to be an integer. "18" is a sensible input value too IMO.

echarrod: It is, I wouldn't have thought Core would handle that request though after they've implemented the changes. 

Additionally, as much as possible we want the SDKs to all behave in the same way. Since we only allow integers in the other SDKs, we should try to make this the same 👍

yoti_python_sdk/tests/dynamic_sharing_service/policy/test_dynamic_policy_builder.py
echarrod: What do you think about this one @emmas-yoti?

yoti_python_sdk/dynamic_sharing_service/dynamic_scenario_builder.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    @param policy A DynamicPolicy defining the attributes to be shared
+    """
+
+    def with_policy(self, policy):
+        self.__scenario["policy"] = policy
+        return self
+
+    """
+    @param extension A Extensions to add to the scenario
```
echarrod:
```suggestion
    @param extension An extension to be activated for the scenario
```

examples/yoti_example_django/yoti_example/views.py
```diff
@@ -1,51 +1,80 @@
...
+            context["age_verified"] = age_verified
 
-        selfie = context.get('selfie')
+        selfie = context.get("selfie")
         if selfie is not None:
             self.save_image(selfie.value)
         return self.render_to_response(context)
 
     @staticmethod
     def save_image(selfie_data):
```
MrBurtyyy: Is there any reason we need to save the image first to be rendered on the application?  Could we not just render the image like ``?

echarrod: We don't have to, but we decided to do this a while ago so that clients can choose between what they have to do. We originally displayed two selfies - one from a saved file, and one in base64. 

Now that we are only displaying one with the new designs, I guess we need to reassess whether it's worth having.

emmas-yoti: I don't really have any strong feelings about this. I can delete it, or I can leave it.

echarrod: For now let's leave, if we choose to remove, we should do so in all of the SDKs

yoti_python_sdk/dynamic_sharing_service/extension/location_constraint_extension_builder.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+
+class LocationConstraintExtensionBuilder(object):
```
MrBurtyyy: Should this extend the ExtensionBuilder rather than being its own standalone class?

```suggestion
from .extension_builder import ExtensionBuilder

class LocationConstraintExtensionBuilder(ExtensionBuilder):
```

yoti_python_sdk/tests/dynamic_sharing_service/policy/test_dynamic_policy_builder.py
```diff
@@ -0,0 +1,119 @@
...
+        .build()
+    )
+
+    assert len(policy["wanted_auth_types"]) == 3
+    assert DynamicPolicyBuilder.SELFIE_AUTH_TYPE in policy["wanted_auth_types"]
+    assert DynamicPolicyBuilder.PIN_AUTH_TYPE in policy["wanted_auth_types"]
+    assert 99 in policy["wanted_auth_types"]
+
+
+def test_build_with_auth_types_true():
```
MrBurtyyy: Not sure if the test is incorrectly named, or if the test doesn't reflect the name of the test?  Should it be:
`policy = DynamicPolicyBuilder().with_selfie_auth(False).build()`
or
`policy = DynamicPolicyBuilder().with_selfie_auth(True).build()`

yoti_python_sdk/dynamic_sharing_service/extension/location_constraint_extension_builder.py
emmas-yoti: I... would say no as that would allow people to change the ExtensionType on a LocationConstraintExtension. Which is wrong.

It's more of an "Has-a"/aggregation than an "Is-a" inheritance relationship.

echarrod: I think Alex was just asking because isn't that what the [TransactionalFlowExtensionBuilder](https://github.com/getyoti/yoti-python-sdk/pull/71/files#diff-6e8884471a7d49961ae6e1d004a72fe5) is doing?

emmas-yoti: Yeah. I wasn't super happy about that, but TransactionalFlow needed the same methods as Extension. I think Extension and TransactionalFlowExtension are... less concrete/well defined? Compared to LocationConstraint.

echarrod: I think we should probably be consistent and have either both of them or none of them using ExtensionBuilder. I'll leave it up to you which one you prefer 👍

yoti_python_sdk/tests/dynamic_sharing_service/policy/test_dynamic_policy_builder.py
echarrod: Marking as resolved, as escaping line breaks isn't very Python-esque

PR #74, opened 2019-07-26
SDK-1031: Add support for Unknown Anchor extensions

emmas-yoti:
### Changed
 * parse_anchors now adds an Unknown anchor if it cannot match an anchor extension against an anchor from the protocol buffer.
 * Anchor.get_values_from_extensions now adds an extension with value "" if an extension does not have a value
 * Replaced test fixtures with those from Ruby
 * This required changing the private key used by the unit tests
 * This in turn required regenerating the signed digests used by the unit tests


yoti_python_sdk/anchor.py
```diff
@@ -1,141 +1,178 @@
...
+        return parsed_anchors
+
+    @staticmethod
+    def get_values_from_extensions(
+        anc, anchor_type, extensions, crypto_cert, parsed_anchors
+    ):
+        if hasattr(extensions, "value"):
+            extension_value = ""
+            if hasattr(extensions.value, "value"):
+                extension_value = Anchor.decode_asn1_value(extensions.value.value)
```
davidgrayston: Will `extensions` always be empty of unknown anchors here or should we skip this when we know it's an unknown _oid_?

yoti_python_sdk/anchor.py
```diff
@@ -1,141 +1,178 @@
...
+
+    @staticmethod
+    def parse_anchors(anchors):
+        if anchors is None:
+            return None
+
+        parsed_anchors = []
+        for anc in anchors:
+            if hasattr(anc, "origin_server_certs"):
+                anchor_type = "UNKNOWN"
```
echarrod: We have SDK-1028 to change this.
Since clients _could_ potentially be using this string to switch, we have to treat it as a breaking change, even though the chance of them doing this is very minimal. We'll have to wait til v3.0.0 to change this. 
```suggestion
                anchor_type = "Unknown"
```

yoti_python_sdk/anchor.py
echarrod: Yes we don't want to parse the unknown anchors since we won't know the encoding scheme in advance. See SDK-621
for further details. We just want to set the unknown anchor's value to `""`

yoti_python_sdk/anchor.py
```diff
@@ -1,141 +1,188 @@
...
+from OpenSSL import crypto
+
+from yoti_python_sdk import config
+
+UNKNOWN_EXTENSION = ""
+SOURCE_EXTENSION = "1.3.6.1.4.1.47127.1.1.1"
+VERIFIER_EXTENSION = "1.3.6.1.4.1.47127.1.1.2"
+
+UNKNOWN_ANCHOR_TYPE = "Unknown"
+UNKNOWN_ANCHOR_SUB_TYPE = "TEST UNKNOWN SUB TYPE"
```
davidgrayston: This should only be in the test (this is the sub type of the sample unknown anchor). The sub type should be available for all anchors.

yoti_python_sdk/anchor.py
```diff
@@ -1,141 +1,188 @@
...
+                                    type(exc).__name__, exc
+                                )
+                            )
+                        continue
+
+                if not has_found_anchor:
+                    parsed_anchors.append(
+                        Anchor(
+                            UNKNOWN_ANCHOR_TYPE,
+                            UNKNOWN_ANCHOR_SUB_TYPE,
```
davidgrayston: Related to previous comment - I think this should be `anc.sub_type`

emmas-yoti:
I'd accidentally checked in .pre-commit-config.yaml in an earlier commit so had to rewrite history to prevent a merge conflict at a future date.

PR #75, opened 2019-07-26
SDK-570: Add remember_me_id to ActivityDetails

emmas-yoti:
### Added
ActivityDetails.remember_me_id property

### Changed
ActivityDetails.user_id attribute changed to property with setter decorator

### Deprecated
ActivityDetails.user_id marked as deprecated.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,8 @@ certifi==2018.11.29       # via requests
 cffi==1.11.5              # via cryptography
 chardet==3.0.4            # via requests
 click==6.6
-colorama==0.3.9           # via pytest
 cryptography==2.4.1
+deprecated==1.2.6
 future==0.15.2
 idna==2.7                 # via cryptography, requests
 itsdangerous==0.24
@@ -30,3 +30,7 @@ six==1.10.0               # via cryptography, mock, protobuf, pyopenssl, pytest
 urllib3==1.24.2
 virtualenv==13.1.2
 wheel==0.24.0
+wrapt==1.11.2             # via deprecated
+
+# The following packages are considered to be unsafe in a requirements file:
+# setuptools==41.0.1        # via protobuf, pytest
--- a/yoti_python_sdk/activity_details.py
+++ b/yoti_python_sdk/activity_details.py
@@ -2,6 +2,7 @@
 import collections
 import json
 import logging
+from deprecated import deprecated
 from datetime import datetime
 
 from yoti_python_sdk import attribute_parser, config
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -1,32 +1,36 @@
...
-#    pip-compile --output-file=requirements.txt requirements.in
-#
-asn1==2.2.0
-asn1crypto==0.24.0        # via cryptography
-attrs==18.1.0             # via pytest
-certifi==2018.11.29       # via requests
-cffi==1.11.5              # via cryptography
-chardet==3.0.4            # via requests
-click==6.6
-colorama==0.3.9           # via pytest
```
davidgrayston: Was this removal expected?

yoti_python_sdk/activity_details.py
```diff
@@ -87,17 +110,23 @@ def try_convert_structured_postal_address_to_dict(field):
...
 
     def __iter__(self):
-        yield 'user_id', self.user_id
-        yield 'parent_remember_me_id', self.parent_remember_me_id
-        yield 'outcome', self.outcome
-        yield 'receipt_id', self.receipt_id
-        yield 'user_profile', self.user_profile
-        yield 'profile', self.profile
-        yield 'base64_selfie_uri', self.base64_selfie_uri
+        yield "user_id", self.user_id
```
davidgrayston: Should `remember_me_id` be added in here?

emmas-yoti: Yes. Yes it should. Ooops.

requirements.txt
emmas-yoti: No. No it was not. D:

I just ran the pip-compile instruction. Not sure what's caused that to be removed.

echarrod: Hmm, yes looks like `pytest` is the same version as before, so don't know why this would have been removed? The only thing that's changed is `pip-compile` version, so maybe it could be more intelligent into what it considers a transitive dependency to be. In any case, I don't think it's something to worry about, and we can leave as is.

yoti_python_sdk/activity_details.py
```diff
@@ -14,26 +15,30 @@ def __init__(self, receipt, decrypted_profile=None):
...
+        if decrypted_profile and hasattr(decrypted_profile, "attributes"):
             decrypted_profile_attributes = decrypted_profile.attributes
             self.profile = Profile(decrypted_profile_attributes)
 
             for field in decrypted_profile_attributes:  # will be removed in v3.0.0
                 try:
                     value = attribute_parser.value_based_on_content_type(
-                        field.value,
-                        field.content_type
+                        field.value, field.content_type
```
echarrod: I'm not a fan of this formatting that `black` does! Are we able to change it?

yoti_python_sdk/activity_details.py
```diff
@@ -14,26 +15,30 @@ def __init__(self, receipt, decrypted_profile=None):
...
                     if field.name == config.ATTRIBUTE_SELFIE:
                         self.base64_selfie_uri = value.base64_content()
-                        value = field.value  # set value to be byte content, for backwards compatibility
+                        value = (
+                            field.value
+                        )  # set value to be byte content, for backwards compatibility
 
                     if field.name == config.ATTRIBUTE_STRUCTURED_POSTAL_ADDRESS:
-                        value = self.try_convert_structured_postal_address_to_dict(field)
+                        value = self.try_convert_structured_postal_address_to_dict(
```
echarrod: Same with `black` here 🤢

echarrod:
Test coverage for `remember_me_id` should just be a case of copying `user_id` checks to also check for `remember_me_id` 👍.

PR #81, opened 2019-08-13
SDK-826: Verify SSL and API Endpoint Environment Variables

MrBurtyyy:
- Add environment variable support to allow SSL verification to be turned off
- Add environment variable support to allow overriding of the whole API URL.

### DIFF
--- a/yoti_python_sdk/__init__.py
+++ b/yoti_python_sdk/__init__.py
@@ -21,6 +21,7 @@ ver_path = convert_path(version_path)
 with open(ver_path) as ver_file:
     exec(ver_file.read(), main_ns)
 
+<<<<<<< HEAD
 __version__ = main_ns["__version__"]
 YOTI_API_URL = environ.get("YOTI_API_URL", DEFAULTS["YOTI_API_URL"])
 YOTI_API_PORT = environ.get("YOTI_API_PORT", DEFAULTS["YOTI_API_PORT"])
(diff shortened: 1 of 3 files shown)


MrBurtyyy:
Fixed the merge conflicts with development

yoti_python_sdk/__init__.py
```diff
@@ -22,10 +23,14 @@
...
-YOTI_API_PORT = environ.get("YOTI_API_PORT", DEFAULTS["YOTI_API_PORT"])
+YOTI_API_PORT = environ.get('YOTI_API_PORT', DEFAULTS["YOTI_API_PORT"])
 YOTI_API_VERSION = environ.get("YOTI_API_VERSION", DEFAULTS["YOTI_API_VERSION"])
-YOTI_API_ENDPOINT = "{0}:{1}/api/{2}".format(
-    YOTI_API_URL, YOTI_API_PORT, YOTI_API_VERSION
-)
+YOTI_API_ENDPOINT = environ.get("YOTI_API_ENDPOINT", "{0}:{1}/api/{2}".format(YOTI_API_URL, YOTI_API_PORT, YOTI_API_VERSION))
+
+YOTI_API_VERIFY_SSL = environ.get("YOTI_API_VERIFY_SSL", DEFAULTS["YOTI_API_VERIFY_SSL"])
+if YOTI_API_VERIFY_SSL == "false" or YOTI_API_VERIFY_SSL == False:
```
echarrod: Although it might be excessive to write a helper function here, I think we need to at least also check for `"False"` if we are to check for `"False"` and `False`

PR #84, opened 2019-08-14
Release 2.8.1

emmas-yoti:
### TO DO
 * [x] Update version numbers (PR#83)
 * [x] Rename methods in WantedAttributeBuilder (PR#86)

### Added
 * `ActivityDetails.remember_me_id` property
 * `ActivityDetails.application_profile` property
 * `ApplicationProfile` class
 * Support for Dynamic QR Codes/Shares with `dynamic_sharing_service` module
 * Document Details support

### Changed
 * Added `Unknown` as an option for the anchor type

### Deprecated
 * `ActivityDetails.user_id` marked as deprecated.


MrBurtyyy:
According to the Java release wiki, changelogs shouldn't contain the following:

Changes to library versions
CVE exclusions
Internal refactorings
Changes to documentation or the demo app

I think there are a few things we can remove from the changelog, I'll add another comment with updates I think we should do.

MrBurtyyy:
### Added
 * `ActivityDetails.remember_me_id` property
 * module `dynamic_sharing_service` with children policy and extension
 * class `dynamic_sharing_service.DynamicScenarioBuilder`
 * classes `DynamicPolicyBuilder` and `WantedAttributeBuilder` in `dynamic_sharing_service.policy`
 * classes `ExtensionBuilder`, `LocationConstraintExtensionBuilder` and `TransactionalFlowExtensionBuilder` in `dynamic_sharing_service.extension`
 * Module `dynamic_sharing_service.share_url` with method `create_share_url`
 * method `make_request` added to `Client`
 * `DocumentDetails` class added
 * `document_details` property added to `Profile` class
 * `document_details` support added to `DynamicPolicyBuilder`
 * Add environment variable support to allow SSL verification to be turned off

### Changed
 * `parse_anchors` now adds an Unknown anchor if it cannot match an anchor extension against an anchor from the protocol buffer.
 * `Anchor.get_values_from_extensions` now adds an extension with value "" if an extension does not have a value
 * `ActivityDetails.user_id` attribute changed to property with setter decorator
 * private `Endpoints` object in `Client` given public getter, endpoints, to be used with make_request

### Deprecated
 * `ActivityDetails.user_id`

yoti_python_sdk/anchor.py
```diff
@@ -1,141 +1,187 @@
...
+VERIFIER_EXTENSION = "1.3.6.1.4.1.47127.1.1.2"
+
+UNKNOWN_ANCHOR_TYPE = "Unknown"
+UNKNOWN_ANCHOR_VALUE = ""
+
+
+class Anchor:
+    def __init__(
+        self,
+        anchor_type="Unknown",
```
echarrod:
```suggestion
        anchor_type=UNKNOWN_ANCHOR_TYPE,
```

echarrod: Actually do we need to initialise it if we're setting it to `Unknown` on line 55?

yoti_python_sdk/client.py
```diff
@@ -79,21 +93,53 @@ def perform_aml_check(self, aml_profile):
...
+        http_method = "POST"
 
         response = self.__make_aml_check_request(http_method, aml_profile)
 
         return aml.AmlResult(response.text)
 
-    def __make_activity_details_request(self, encrypted_request_token, http_method, content):
-        decrypted_token = self.__crypto.decrypt_token(encrypted_request_token).decode('utf-8')
+    def make_request(self, http_method, endpoint, body):
+
```
echarrod:
```suggestion
```

yoti_python_sdk/anchor.py
emmas-yoti: My preference it to initialise it in case a customer uses it in a weird way.

echarrod: So we could just initialise it here, and then we don't need to on L55? It's safer to do both, but I'm not sure if it's necessary. I'm easy either way

MrBurtyyy:
Updated changelog to include ApplicationProfile changes which were merged in last minute

yoti_python_sdk/anchor.py
emmas-yoti: ¯\_(ツ)_/¯

PR #85, opened 2019-08-14
SDK-1098: Add support for Application Profile

MrBurtyyy:
- Added class `ApplicationProfile` and updated `Profile` which both inherit from new class `BaseProfile`.
- Updated `ActivityDetails` to accept decrypted application profile attributes
- Updated tests for profiles to include tests for `ApplicationProfile`


yoti_python_sdk/client.py
```diff
@@ -78,16 +78,26 @@ def get_activity_details(self, encrypted_request_token):
...
-        return ActivityDetails(receipt, attribute_list)
+        decrypted_application_data = self.__crypto.decipher(
+            unwrapped_key, encrypted_application_profile.iv, encrypted_application_profile.cipher_text
+        )
+
+        user_profile_attribute_list = proto.attribute_list(decrypted_profile_data)
+        application_profile_attribute_list = proto.attribute_list(decrypted_application_data)
+
+        return ActivityDetails(
+            receipt, user_profile_attribute_list, decrypted_application_profile=application_profile_attribute_list
```
echarrod: I prefer taking the approach that we have _all_ or _0_ named arguments, so I'll leave it up to you whether we name the other arguments, or take out the `decrypted_application_profile=`

PR #90, opened 2019-08-21
SDK-1123: source constraints

emmas-yoti:
### Additions
 * Add SourceConstraintBuilder
 * Add WantedAnchorBuilder

### Changed
 * Support for source constraints in WantedAttributeBuilder
 * Support for source constraints in DynamicPolicyBuilder
 * 'optional' field to WantedAttributeBuilder
 * 'accept_self_asserted' field to WantedAttributeBuilder
 * 'auto_allow' field in DynamicScenarioBuilder


yoti_python_sdk/tests/dynamic_sharing_service/test_dynamic_scenario_builder.py
```diff
@@ -31,3 +31,14 @@ def test_build_scenario():
     assert EXTENSION1 in scenario["extensions"]
     assert EXTENSION2 in scenario["extensions"]
     assert scenario["callback_endpoint"] == CALLBACK_ENDPOINT
+
+
+def test_auto_allow():
```
MrBurtyyy: If we're going to remove withAutoAllow from the source constraint, we won't need this test anymore

yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
```diff
@@ -24,8 +25,35 @@ def with_derivation(self, derivation):
...
+            self.__constraints.append(constraint)
+        return self
+
     def build(self):
         """
         :return: The wanted attribute object
         """
-        return self.__attribute.copy()
+        attribute = self.__attribute.copy()
+        attribute["constraints"] = self.__constraints[:]
```
davidgrayston: @echarrod Should constraints only be included in the payload if they have been added with the builder?

It shouldn't matter if an empty array is provided, so should be fine like this, but wondered if this is something we should keep consistent across SDKs?

echarrod: Ideally yes we'd keep it so it behaves as it currently does without constraints i.e. not including constraints in the payload if they're not specified 👍

yoti_python_sdk/config.py
```diff
@@ -30,3 +30,7 @@
 X_YOTI_SDK = "X-Yoti-SDK"
 X_YOTI_SDK_VERSION = X_YOTI_SDK + "-Version"
 JSON_CONTENT_TYPE = "application/json"
+PASSPORT = "PASSPORT"
```
echarrod: Could we change these to have something like `ANCHOR_VALUE_` in front of them, just so it's evident from just so they have a bit more context 👍 
```suggestion
ANCHOR_VALUE_PASSPORT = "PASSPORT"
```

yoti_python_sdk/dynamic_sharing_service/dynamic_scenario_builder.py
```diff
@@ -12,33 +12,31 @@ def __init__(self):
             "callback_endpoint": "",
         }
 
-    """
-    @param policy A DynamicPolicy defining the attributes to be shared
-    """
-
     def with_policy(self, policy):
+        """
+        :param policy: A DynamicPolicy defining the attributes to be shared
```
echarrod: I know this was here previously, but technically it's the attributes to be _requested_. Small difference, but with soft preferences, and potentially optional attributes in future, it's an important distinction 👍 
```suggestion
        :param policy: A DynamicPolicy defining the attributes to be requested
```

yoti_python_sdk/dynamic_sharing_service/policy/source_constraint_builder.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        :param value: If true, this will be treated as a soft preference, otherwise
+                      this will be treated as a hard requirement
+        """
+        self.__soft_preference = value
+        return self
+
+    def with_anchor(self, anchor):
+        """
+        :param anchor: Add an anchor to the source constraint. You should prefer
```
echarrod: Not sure if we should say `You should prefer (to) use`, should we say `It is recommended to use the other helper methods instead of this one`?

yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
```diff
@@ -24,8 +25,35 @@ def with_derivation(self, derivation):
...
         return self
 
+    def with_accept_self_asserted(self, value=True):
+        """
+        :param value: True if self-asserted details are allowed
+        """
+        self.__attribute["accept_self_asserted"] = value
+        return self
+
+    def with_optional(self, value=True):
```
echarrod: This is not supported by the app yet, so we need to remove.
In the other SDKs I believe we've kept it in the payload, but we're always setting it to false.

yoti_python_sdk/dynamic_sharing_service/policy/source_constraint_builder.py
```diff
@@ -0,0 +1,77 @@
...
+        :param value: If true, this will be treated as a soft preference, otherwise
+                      this will be treated as a hard requirement
+        """
+        self.__soft_preference = value
+        return self
+
+    def with_anchor(self, anchor):
+        """
+        :param anchor: Add an anchor to the source constraint.
+        It is recommended to use the other helper methods isntead of this one
```
echarrod:
```suggestion
        It is recommended to use the other helper methods instead of this one
```

PR #91, opened 2019-08-21
SDK-1160: Fix anchor parsing critical extensions

MrBurtyyy:
- Fix anchor parser to stop the parsing of critical extensions after one non-critical extension has been parsed


yoti_python_sdk/anchor.py
```diff
@@ -72,6 +72,8 @@ def parse_anchors(anchors):
 
                 has_found_anchor = False
                 for i in range(len(crypto_cert.extensions)):
+                    anchor_type = UNKNOWN_ANCHOR_TYPE
```
davidgrayston: Could we remove `anchor_type = UNKNOWN_ANCHOR_TYPE` further up in this method?

PR #93, opened 2019-08-21
Release 2.8.2: Bump version number

MrBurtyyy:
- Bumped the version number to 2.8.2 for release

### DIFF
--- a/yoti_python_sdk/version.py
+++ b/yoti_python_sdk/version.py
@@ -1,2 +1,2 @@
 # -*- coding: utf-8 -*-
-__version__ = "2.8.1"
+__version__ = "2.8.2"


yoti_python_sdk/version.py
```diff
@@ -1,2 +1,2 @@
 # -*- coding: utf-8 -*-
-__version__ = "2.8.1"
+__version__ = "2.8.2"
```
emmas-yoti: We should update the requirements.{in,txt} files as well I think?

echarrod: We can do that after it's released to PyPi - as `pip-compile` checks to see if the package is live before generating the `requirements.txt`

PR #94, opened 2019-08-22
SDK-1161: Update Flask dependency

MrBurtyyy:
- Update requirements.in file for example Flask project to require Flask>=1.0.4
- Update requirements.in file for example Flask project to require Yoti>=2.8.1

### DIFF
--- a/examples/yoti_example_flask/requirements.txt
+++ b/examples/yoti_example_flask/requirements.txt
@@ -11,7 +11,7 @@ cffi==1.11.5              # via cryptography
 chardet==3.0.4            # via requests
 click==6.7                # via flask
 cryptography==2.5
-flask==1.0.2
+flask==1.1.1
 future==0.16.0            # via yoti
 idna==2.7                 # via requests
 itsdangerous==0.24        # via flask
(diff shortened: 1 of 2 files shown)


MrBurtyyy:
I had to update the Yoti SDK version as they were pointing at 2.8.0 which doesn't exist, so therefore couldn't compile the requirements.txt file

examples/yoti_example_flask/requirements.txt
```diff
@@ -24,8 +24,8 @@ python-dotenv==0.8.2
 requests==2.21.0
 six==1.11.0               # via cryptography, protobuf, pyopenssl
 urllib3==1.24.2
-werkzeug==0.14.1          # via flask
-yoti==2.8.0
+werkzeug==0.15.5          # via flask
+yoti==2.8.1
```
emmas-yoti: Shouldn't this be 2.8.2 now?

MrBurtyyy: This can't be updated until 2.8.2 has been released on PyPi so once we have released it, we will update the example projects to use 2.8.2

PR #98, opened 2019-08-22
SDK-573: Implement AgeVerification

MrBurtyyy:
## Added

* `AgeVerification` class
  * `age` property - the requested age the date of birth should be over/under
  * `check_type` property - the type of check performed (age over/under)
  * `result` property - the result of the check (True/False)
  * `attribute` property - the attribute the age verification was derived from
* `Profile` class
  * `verifications` property - a dictionary of all the age verifications
  * `get_age_verifications()` - returns a list of all the age verifications
  * `find_age_over_verification(age_over)` - returns a matching `AgeVerification` based on `age_over` parameter
  * `find_age_under_verification(age_under)` - returns a matching `AgeVerification` based on `age_under` parameter

## Deprecated

* `try_parse_age_verified_field` method on `ActivityDetails` class

### DIFF
--- a/yoti_python_sdk/profile.py
+++ b/yoti_python_sdk/profile.py
@@ -176,7 +176,8 @@ class Profile(BaseProfile):
     def __find_all_age_verifications(self):
         if self.verifications is None:
             self.verifications = {}
-            for attribute in self.attributes:
+            for key in self.attributes:
+                attribute = self.attributes[key]
                 if (
                     config.ATTRIBUTE_AGE_OVER in attribute.name
                     or config.ATTRIBUTE_AGE_UNDER in attribute.name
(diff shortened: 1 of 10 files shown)


yoti_python_sdk/profile.py
```diff
@@ -161,6 +161,27 @@ def document_images(self):
     def document_details(self):
         return self.get_attribute(config.ATTRIBUTE_DOCUMENT_DETAILS)
 
+    def get_age_verifications(self):
```
davidgrayston: Can we add some test coverage for these new methods?

MrBurtyyy: Added them [here](https://github.com/getyoti/yoti-python-sdk/pull/98/commits/336489b9d561540e0afacb95c40b42224b461bac)

MrBurtyyy:
I've updated the example projects to use the new AgeVerification objects, so if either @davidgrayston or @echarrod could take a quick look and if all good I will merge to development 👍

PR #99, opened 2019-08-23
SDK-1132: Sonaqube analysis

MrBurtyyy:
## Added

* Pytest and pytest-cov coverage reports
* Sonaqube properties file
* Pylint for code smells

PR #100, opened 2019-08-23
SDK-602: Sandbox support

MrBurtyyy:
- [x] Merge #99 into development

This PR contains all of the relevant supporting objects for the sandbox integrations tests.

## Added

* New module `yoti_python_sdk.sandbox`
* `SandboxClient` class
* `YotiTokenRequest` class
* `YotiTokenResponse` class
* `SandboxAttribute` class
* `SandboxAnchor` class
* `SandboxAgeVerification` class


yoti_python_sdk/sandbox/anchor.py
```diff
@@ -0,0 +1,78 @@
+UNKNOWN_EXTENSION = ""
+SOURCE_EXTENSION = "1.3.6.1.4.1.47127.1.1.1"
+VERIFIER_EXTENSION = "1.3.6.1.4.1.47127.1.1.2"
+
+UNKNOWN_ANCHOR_TYPE = "Unknown"
+UNKNOWN_ANCHOR_VALUE = ""
+
```
echarrod: Was going to say that we can just use the constants in `yoti_python_sdk/anchor` instead of these, but it looks like only `UNKNOWN_ANCHOR_TYPE` is being used in this class (unless it's in the integration tests repo?)

yoti_python_sdk/sandbox/client.py
```diff
@@ -0,0 +1,184 @@
...
+        if isinstance(o, SandboxAttribute):
+            return o.__dict__()
+        if isinstance(o, SandboxAnchor):
+            return o.__dict__()
+
+        return json.JSONEncoder.default(self, o)
+
+
+class SandboxClient(object):
+
```
echarrod: We shouldn't have a blank line after class declarations, guessing this isn't picked up by any of our pre-commit checks?
```suggestion
```

yoti_python_sdk/sandbox/client.py
```diff
@@ -0,0 +1,184 @@
...
+        except (AttributeError, IOError, TypeError, OSError) as exc:
+            error = 'Could not read private key file: "{0}", passed as: {1} '.format(
+                key_file_path, error_source
+            )
+            exception = "{0}: {1}".format(type(exc).__name__, exc)
+            raise RuntimeError("{0}: {1}".format(error, exception))
+
+    @staticmethod
+    def __create_request(http_method, path, content):
+        if http_method not in SandboxClient.HTTP_SUPPORTED_METHODS:
```
echarrod: We could change this to use the methods declared in the SDK couldn't we, to avoid declaring it twice? And we can get rid of the declaration in this class
```suggestion
        if http_method not in yoti_python_sdk.client.HTTP_SUPPORTED_METHODS
```

PR #102, opened 2019-08-23
SDK-1076: Signed Requests

MrBurtyyy:
- [x] Merge #100 into development

### `yoti_python_sdk.http`

* `SignedRequestBuilder`
  * The `SignedRequestBuilder` is used for building a `SignedRequest` which contains all of the required functionality for signing a request to Yoti API services.  This includes building the correct signature headers.
* `SignedRequest`
  * The `SignedRequest` object is what is built after using the `SignedRequestBuilder`.  This object can be used for executing the request using a default (or custom) `RequestHandler`.  The `DefaultRequestHandler` (which uses the requests library) is used by default if no `RequestHandler` is supplied.
  * The `SignedRequest` object also has getter methods to allow you to perform your own HTTP request outside of the `RequestHandler` framework.
* `RequestHandler`
  * The `RequestHandler` interface can be used by the `SignedRequest` when calling the `execute()` function.  If using a custom `RequestHandler`, you must return a `YotiResponse` object.
* `YotiResponse`
  * This class is used internally by the SDK for parsing responses from `SignedRequest`.

### `yoti_python_sdk.utils`

* `create_nonce()`
  * Creates a nonce used for cryptographic functions
* `create_timestamp()`
  * Creates a timestamp as an integer (using Unix epoch time)

### `yoti_python_sdk.endpoint`

* All functions in the `endpoint` module have had an optional parameter added to them which stops query parameters being added to the string.  This is due to the `SignedRequestBuilder` taking precedence over this.

### `yoti_python_sdk.crypto`

* `Crypto`
  * Moved method for reading a PEM file from the `Client` to the `Crypto` class as a static function.

## `yoti_python_sdk.client`

* `Client` 
  * This class now has an optional constructor argument which allows the overriding of the default `RequestHandler`, should a client not want to use the `DefaultRequestHandler` which uses the requests library.
  * The custom request handler you supply MUST return an instance of `YotiResponse`, or the SDK will error due to invalid properties.
* `Client` has been updated to internally use the new `SignedRequestBuilder` and `SignedRequest` for making Yoti API service calls.


MrBurtyyy:
This branch was rebased on #100 due to requiring Sandbox components for testing

yoti_python_sdk/endpoint.py
```diff
@@ -6,30 +7,45 @@ class Endpoint(object):
     def __init__(self, sdk_id):
         self.sdk_id = sdk_id
 
-    def get_activity_details_request_path(self, decrypted_request_token):
+    def get_activity_details_request_path(
+        self, decrypted_request_token, no_params=False
+    ):
+        # TODO: Remove no_params in 3.0.0 and get rid of the query parameters in the formatted
```
emmas-yoti: I'm not sure we should be checking in TODO tags into Git (if we can help it)
We should probably use Jira releases to tack this sort of stuff?

yoti_python_sdk/sandbox/token.py
```diff
@@ -0,0 +1,128 @@
...
+        self.sandbox_attributes = sandbox_attributes
+
+    def __dict__(self):
+        return {
+            "remember_me_id": self.remember_me_id,
+            "profile_attributes": self.sandbox_attributes,
+        }
+
+    @staticmethod
+    def builder():
```
emmas-yoti: As a matter of code quality, do we want to require doc comments on public methods?

yoti_python_sdk/endpoint.py
MrBurtyyy: Yep probably - I'll fix this now

yoti_python_sdk/sandbox/token.py
MrBurtyyy: This is a dependency from #100 so I'll go through all of the Sandbox stuff in that PR and add docstrings, and then once that PR has been merged I'll rebase this branch on development

MrBurtyyy: Fixed in #100

PR #105, opened 2019-09-23
SDK-1196: Correct mispelling of longitude in Python SDK

emmas-yoti:
### Changed
 * Fix broken `location_constraint_extention`, misspelling of longitude causes invalid payload to be generated.

### DIFF
--- a/yoti_python_sdk/tests/dynamic_sharing_service/extension/test_location_constraint_extension_builder.py
+++ b/yoti_python_sdk/tests/dynamic_sharing_service/extension/test_location_constraint_extension_builder.py
@@ -8,14 +8,14 @@ from yoti_python_sdk.dynamic_sharing_service.extension.location_constraint_exten
 
 def test_builds_with_given_values():
     LATITUDE = 50
-    LONGTITUDE = 99
+    LONGITUDE = 99
     RADIUS = 60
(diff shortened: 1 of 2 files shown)


yoti_python_sdk/tests/dynamic_sharing_service/extension/test_location_constraint_extension_builder.py
```diff
@@ -4,18 +4,73 @@
 from yoti_python_sdk.dynamic_sharing_service.extension.location_constraint_extension_builder import (
     LocationConstraintExtensionBuilder,
 )
+import pytest
+
+
+def test_longitude_validation():
```
echarrod: Not essential for right now, but we can use pytest to parametrize all of these new tests, which will make them more concise (like in [test_profile.py](https://github.com/getyoti/yoti-python-sdk/blob/development/yoti_python_sdk/tests/test_profile.py#L162))
Something like:
```
@pytest.mark.parametrize(
    "longitude",
    [
        270.00,
        -181,
        "27.00",
    ],
)
...assert raises error

@pytest.mark.parametrize(
    "longitude",
    [
        180.0,
        -90,
    ],
)
...assert doesn't raise error
```

PR #110, opened 2019-10-02
SDK-1206: Remove X-Auth-Key from default headers of request builder

emmas-yoti:
### Changed
 * `X-Auth-Key` removed from `SignedRequest`'s default headers
 * `X-Auth-Key` being set by business logic in `Client`


emmas-yoti:
> +1 can we add coverage for this header in `test_client.py`?

Have had a 

> +1 can we add coverage for this header in `test_client.py`?

Have had an offline conversation. It's already covered by the existing test fixtures.

PR #111, opened 2019-10-10
SDK-1241: Add coveralls support

MrBurtyyy:
* Update the travis file to include coveralls support


requirements.txt
```diff
@@ -31,10 +32,11 @@ py==1.5.3                 # via pytest
 pycparser==2.18           # via cffi
 pyopenssl==18.0.0
 pytest==3.3.2
+python-coveralls==2.9.3
```
davidgrayston: Is it possible to move dev dependencies (such as `python-coveralls`, `mock`, `pytest`) into a separate file, e.g. `requirements-dev.txt`? (Probably a separate task if we decide to do that)

echarrod: I think you maybe put these in the `extras_require={` part of the setup.py file? Not sure though

MrBurtyyy: I've removed all the development dependencies out of requirements.in, and into the extras_require section of setup.py as @echarrod suggested.  Now to install the SDK for development, we would run:

```bash
pip install -r requirements.txt
pip install -e .[dev]
```

PR #114, opened 2019-10-29
SDK-1229: Attribute Issuance Details

emmas-yoti:
### Additions
 * `date_parser.datetime_with_microseconds` implements a simple, fairly permissive, parser for iso dates that expects and enforces microsecond accuracy. This is a workaround for Python 2.7 which will soon EOL and is expected to be deprecated almost immediately.
 * `IssuanceDetails` and `ExtraData` to represent parsed Extra Data and Thirdparty Attributes
 * `extra_data` and `thirdparty_attribute` functions to recursively parse extra data


emmas-yoti:
There's a build failure on 3.8. We've discussed this and we think this might be an upstream issue with Protobufs on 3.8 that we're expecting to resolve once 3.8 is more than a week old.

yoti_python_sdk/tests/share/test_extra_data.py
```diff
@@ -0,0 +1,122 @@
...
+    result = extra_data.attribute_issuance_details
+    assert result is not None
+    assert result.token == "someIssuanceToken"
+    assert result.expiry_date == datetime(2019, 10, 15, 22, 4, 5, 123000)
+    assert result.attributes[0].name == "com.thirdparty.id"
+    assert result.attributes[1].name == "com.thirdparty.other_id"
+
+
+def test_should_handle_no_expiry_date():
+    tokenValue = "tokenValue"
```
MrBurtyyy: Can we add in a test to make sure `attribute_issuance_details` on extra data should be `None` when there is no token?

emmas-yoti: test_attribute_issuance_details_should_return_nil_when_no_data_entries.

Will rename s/nil/none/ so it greps easier.

yoti_python_sdk/tests/test_issuance_details.py
```diff
@@ -0,0 +1,50 @@
...
+def test_should_parse_third_party_attribute_correctly():
+    thirdparty_attribute_bytes = file_helper.get_file_bytes(THIRD_PARTY_ATTRIBUTE)
+
+    proto = ThirdPartyAttribute_pb2.ThirdPartyAttribute()
+    proto.MergeFromString(thirdparty_attribute_bytes)
+
+    issuance_details = IssuanceDetails(proto)
+
+    assert issuance_details.attributes[0].name == "com.thirdparty.id"
+    assert issuance_details.token == "someIssuanceToken"
```
davidgrayston: This should be base64 encoded: `29tZUlzc3VhbmNlVG9rZW4=`

yoti_python_sdk/date_parser.py
```diff
@@ -9,4 +14,29 @@ def from_iso_format(string):
...
+    # Python2 does not have a way of parsing date formats.
+    # TODO Deprecate this once Python2 support is dropped.
+    time_split = re.split("[^0-9]", string)
+    parts = len(time_split)
+    if parts <= 6:
+        if logging.getLogger().propagate:
+            logging.warning(ERROR_PARSING_DATE)
+        return None
+
+    year = int(time_split[0]) if parts > 0 else 0
```
echarrod: Not sure if I'm keen on how this is structured - surely we don't need to check for parts <= 6 here if we're checking on line 25? As long as we have a test covering that scenario, any other checking is just adding noise

yoti_python_sdk/date_parser.py
```diff
@@ -9,4 +14,29 @@ def from_iso_format(string):
...
+    hour = int(time_split[3]) if parts > 3 else 0
+    minute = int(time_split[4]) if parts > 4 else 0
+    second = int(time_split[5]) if parts > 5 else 0
+    microsecond = int(float("0." + time_split[6]) * 1e6) if parts > 6 else 0
+    try:
+        return datetime(year, month, day, hour, minute, second, microsecond)
+    except ValueError:
+        if logging.getLogger().propagate:
+            logging.warning(ERROR_PARSING_DATE)
+        return None
```
echarrod: We need to cover every branch in our code, even when it's an interim solution like this for when we remove Python 2.7 support. 

So because this is not a simple function, can we add tests that cover every possible scenario here? Including:
1. less than 6 parts (but not empty string)
1. that the other parts are set to 0 for each of the differing number of parts
1. That ValueError is caught
1. That other errors are propagated

(Apologies if there are some tests that I'm missing)

yoti_python_sdk/date_parser.py
```diff
@@ -9,4 +14,29 @@ def from_iso_format(string):
     if len(parts) != 3:
         raise ValueError
 
-    return datetime.date(parts[0], parts[1], parts[2])
+    return date(parts[0], parts[1], parts[2])
+
+
+def datetime_with_microsecond(string):
+    # Python2 does not have a way of parsing date formats.
+    # TODO Deprecate this once Python2 support is dropped.
```
echarrod: I prefer not having TODOs in code - we currently don't check for TODOs when releasing, but we do check that all issues in a particular release are done. Can we instead create an issue and add it to Python 3.0.0?
```suggestion
```

yoti_python_sdk/date_parser.py
emmas-yoti: It's because the above check only logs a warning - it doesn't hard fail.

yoti_python_sdk/issuance_details.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
+from yoti_python_sdk import date_parser
+import base64
+
+
+class IssuanceDetails(object):
```
davidgrayston: Should this be `AttributeIssuanceDetails`?

yoti_python_sdk/issuance_details.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, data_entry):
+        self.__token = base64.b64encode(data_entry.issuance_token)
+        self.__expiry_date = date_parser.datetime_with_microsecond(
+            data_entry.issuing_attributes.expiry_date
+        )
+        self.__attributes = data_entry.issuing_attributes.definitions
```
davidgrayston: `__attributes` is `<class 'google.protobuf.pyext._message.RepeatedCompositeContainer'>`

Should this instead be a list of `AttributeDefinition` (or list of dictionary `{"name": "someName"}`)?

davidgrayston: I can see this is now `<class 'list'>` of `IssuingAttributes_pb2.Definition` - should this be a list of dictionary?

yoti_python_sdk/attribute_issuance_details.py
```diff
@@ -0,0 +1,25 @@
...
+import base64
+
+
+class AttributeIssuanceDetails(object):
+    def __init__(self, data_entry):
+        self.__token = base64.b64encode(data_entry.issuance_token)
+        self.__expiry_date = date_parser.datetime_with_microsecond(
+            data_entry.issuing_attributes.expiry_date
+        )
+        self.__attributes = [a for a in data_entry.issuing_attributes.definitions]
```
davidgrayston: I think this needs to be list of dictionary (instead of `IssuingAttributes_pb2.Definition`)

yoti_python_sdk/issuance_details.py
emmas-yoti: Oh. You're right. I just checked the type of the first :(

PR #116, opened 2019-11-04
SDK-1149: Support multiple attributes with the same name

MrBurtyyy:
## Added

* `BaseProfile` now has a property called `attribute_list` which is what the methods now reference in order to get Attributes back from the profile.
* `attributes` is still available for backwards compatibility, but will be removed in the next major.

## Deprecated

* `Profile.attributes` in favour of `Profile.attribute_list`

### DIFF
--- a/yoti_python_sdk/tests/test_profile.py
+++ b/yoti_python_sdk/tests/test_profile.py
@@ -513,8 +513,8 @@ def test_nested_multi_value():
     outer_tuple = (inner_multi_value,)
 
     profile = Profile(profile_attributes=None)
-    profile.attributes[attribute_name] = Attribute(
-        name=attribute_name, value=outer_tuple, anchors=None
+    profile.attribute_list.append(
+        Attribute(name=attribute_name, value=outer_tuple, anchors=None)
     )
 
     retrieved_multi_value = profile.get_attribute(attribute_name)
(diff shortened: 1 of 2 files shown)


MrBurtyyy:
> It looks like we also need to key the wanted attributes on `name + constraints` to allow retrieval of the same attribute with different constraints

Fixed in https://github.com/getyoti/yoti-python-sdk/pull/116/commits/555a589b30291c2f54b4c0712e0471304c5a4896

MrBurtyyy:
@davidgrayston @echarrod @emmas-yoti I have updated the profile to use a dictionary based approach rather than a list like we had originally implemented.

yoti_python_sdk/tests/test_profile.py
```diff
@@ -753,3 +754,57 @@ def test_expect_none_when_no_age_under_verification_exists():
...
+    )
+    second_attr = ProtobufAttribute(
+        "full_name", "some_other_value".encode(), None, Protobuf.CT_STRING
+    )
+    attr_list = [first_attr, second_attr]
+
+    human_profile = Profile(attr_list)
+
+    attribute = human_profile.get_attribute("full_name")
+    assert attribute is not None
```
echarrod: Is this line redundant is we have the line below?

echarrod:
👀

PR #123, opened 2019-11-25
SDK-1281: Prevent Empty Attribute Names in Wanted Attribute

emmas-yoti:
### Changed
 * Add assertions to WantedAttributeBuilder.build to check that the attribute name is set and not-empty.

### DIFF
--- a/yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
+++ b/yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
@@ -8,7 +8,7 @@ class WantedAttributeBuilder(object):
     """
 
     def __init__(self):
-        self.__attribute = {"name": None}
+        self.__attribute = {}
         self.__constraints = []
 
     def with_name(self, name):
@@ -47,7 +47,7 @@ class WantedAttributeBuilder(object):
         """
         :return: The wanted attribute object
         """
-        if self.__attribute["name"] is None or self.__attribute["name"] == "":
+        if self.__attribute.get("name", None) is None or self.__attribute["name"] == "":
             raise ValueError
         attribute = self.__attribute.copy()
         attribute["constraints"] = self.__constraints[:]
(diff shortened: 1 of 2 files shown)


yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
```diff
@@ -47,6 +47,8 @@ def build(self):
         """
         :return: The wanted attribute object
         """
+        assert self.__attribute["name"] is not None
+        assert self.__attribute["name"] != ""
```
MrBurtyyy: I don't think we should be using assert in non-test code.  Something like a `ValueError` would be better, as it's something that a user could recover from.  `assert` is generally used to describe something that should never happen and is non-recoverable.

yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
```diff
@@ -47,8 +47,8 @@ def build(self):
         """
         :return: The wanted attribute object
         """
-        assert self.__attribute["name"] is not None
-        assert self.__attribute["name"] != ""
+        if self.__attribute["name"] is None or self.__attribute["name"] == "":
+            raise ValueError
```
MrBurtyyy: Can we add a test for this?

yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
```diff
@@ -47,6 +47,8 @@ def build(self):
         """
         :return: The wanted attribute object
         """
+        if self.__attribute["name"] is None or self.__attribute["name"] == "":
```
MrBurtyyy:
```suggestion
        if self.__attribute.get("name", None) is None or self.__attribute["name"] == "":
```

yoti_python_sdk/dynamic_sharing_service/policy/wanted_attribute_builder.py
```diff
@@ -8,7 +8,7 @@ class WantedAttributeBuilder(object):
     """
 
     def __init__(self):
-        self.__attribute = {}
+        self.__attribute = {"name": None}
```
MrBurtyyy: I think it's better that we don't default name to None here, but when we try to get it in the build method, set the default value to `None` if the key doesn't exist (see other comment)

```suggestion
        self.__attribute = {}
```

PR #137, opened 2019-12-11
SDK-1337: Update pylint configuration

MrBurtyyy:
## Changed

* Updated pylint configuration to correctly ignore certain directories (such as auto-generated protobuf files), as well as to ignore specific pylint rules that may conflict with black formatter.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,2 +1,3 @@
 [MASTER]
-ignore=yoti_python_sdk/tests/**,examples/**
+ignore=tests,protobuf
+disable=C0330
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,8 @@ setup(
             "pre-commit==1.17.0",
             "pytest>=3.6.0",
             "pytest-cov>=2.7.1",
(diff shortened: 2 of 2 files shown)


.pylintrc
```diff
@@ -1,2 +1,3 @@
 [MASTER]
-ignore=yoti_python_sdk/tests/**,examples/**
\ No newline at end of file
+ignore=tests,examples,protobuf
```
echarrod: Would it be better to qualify these with `yoti_python_sdk/` first, just in case we create any other folders matching these names that we might want to format? (unlikely, i know)

MrBurtyyy: Pylint configuration is a bit janky in the fact that you can't specify a path for a directory you want to ignore.  Instead you have to specify a base name which is part of the module you are linting i.e. to ignore `yoti_python_sdk/tests` I would exclude `tests` and then run `pylint yoti_python_sdk` and it would know to exclude it.

echarrod: Ok, let's leave it as is then 👍

PR #141, opened 2020-01-09
SDK-1364: Ensure datetime parsing can handle timezone and microsecond

emmas-yoti:


yoti_python_sdk/tests/test_date_parser.py
```diff
@@ -179,3 +180,48 @@ def test_parse_timestamp_from_microseconds(input_value, expected_value):
...
+        ),
+        (
+            "2006-01-02T22:04:05.12Z",
+            datetime(2006, 1, 2, 22, 4, 5, int(120e3), pytz.utc),
+        ),
+        (
+            "2006-01-02T22:04:05.123Z",
+            datetime(2006, 1, 2, 22, 4, 5, int(123e3), pytz.utc),
+        ),
+        ("2006-01-02T22:04:05.1234Z", datetime(2006, 1, 2, 22, 4, 5, 123400, pytz.utc)),
```
echarrod: Annoying that black changes some of these to be split onto two lines, and not others 😒 
I guess it can't be helped if we stick to black...

PR #148, opened 2020-02-11
SDK-1443: Remove regex validation from document details

emmas-yoti:
Support for parsing redacted aadhaar numbers

### Changed
 * Removed regex validation from `DocumentDetails`

### DIFF
--- a/yoti_python_sdk/document_details.py
+++ b/yoti_python_sdk/document_details.py
@@ -27,8 +27,8 @@ class DocumentDetails(object):
         return self.__dict__.get("_DocumentDetails__issuing_authority", None)
 
     def __parse_data(self, data):
-        data = data.split()
-        if len(data) < 3:
+        data = data.split(" ")
+        if len(data) < 3 or "" in data:
             raise ValueError("Invalid value for DocumentDetails")
 
         self.__document_type = data[0]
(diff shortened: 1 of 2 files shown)


yoti_python_sdk/document_details.py
```diff
@@ -31,14 +26,10 @@ def expiration_date(self):
...
-    def __validate_data(self, data):
-        if self.VALIDATION_REGEX.search(data):
-            return
-        else:
-            raise ValueError("Invalid value for DocumentDetails")
-
     def __parse_data(self, data):
         data = data.split()
+        if len(data) < 3:
+            raise ValueError("Invalid value for DocumentDetails")
```
MrBurtyyy: Let's check for empty strings in the list (as this could be indicative of consecutive spaces in the document details string) and add some tests for it (the other SDK's should have examples for this)

emmas-yoti: test_should_parse_with_double_space (test_document_details.py:127) covers this test case. Apparently the implementation of string split on python makes this impossible.

emmas-yoti: The implementation of string split() in Python swallows this. Have added test_should_fail_with_double_space to demonstrate this.

emmas-yoti:
> Still missing the logic for spaces in the list once the string has been split, as well as tests for it (which the other SDK's should have as well)

Ahhh. You replied before I could fix it. Sorry.

PR #154, opened 2020-02-21
Release 2.10.1

emmas-yoti:
### Changed
 * Document Details Attribute to support redacted document numbers, such as an Aadhaar number.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ git:
   depth: 1
 
 jobs:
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -4,9 +4,11 @@ dist: xenial
   
 # Only clone the most recent commit.
```
davidgrayston: We should probably remove/update this comment

PR #156, opened 2020-03-06
SDK-1322: Update dependabot reviewers

echarrod:


MrBurtyyy (approved):

davidgrayston (approved):

echarrod:
Closing as this is rebased on development, will reopen other one

mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

Diff reviewed: `git diff origin/development...HEAD` on branch `websdk-auto/SDK-2765-python-support-handled-check-limit-property-for-checks-in-idv-sandbox`.

---

### Critical

None

---

### Major

**1. `yoti_python_sdk/doc_scan/session/create/subcheck/sub_check.py:20` — `SubRequestedCheck.to_json()` accesses `self.config` which is undefined and not declared abstract**

`to_json()` references `self.config` but the class defines only `type` as `@abstractmethod`. There is no `config` property defined or declared abstract. Any subclass that implements `type` but not `config` will silently compile and crash at runtime with `AttributeError` when serialised.

Verified by running:
```python
class BadSubCheck(SubRequestedCheck):
    @property
    def type(self): return 'SOME_TYPE'
    # no config defined

b = BadSubCheck()
b.to_json()  # → AttributeError: 'BadSubCheck' object has no attribute 'config'
```

Suggested fix — mirror the pattern in `RequestedCheck` and declare `config` abstract:
```python
@property
@abstractmethod
def config(self):
    """Return the configuration for the sub-check."""
    raise NotImplementedError
```

---

### Minor

**2. `yoti_python_sdk/doc_scan/session/create/subcheck/sub_check.py:12` — Python 2-style `__metaclass__ = ABCMeta` is inoperative in Python 3**

The project now targets Python 3.9–3.12 exclusively (updated CI matrix). The `__metaclass__ = ABCMeta` assignment is silently ignored by Python 3; the `@abstractmethod` decorator has no enforcement effect, and `SubRequestedCheck` can be instantiated without implementing any abstract method. This gives false safety guarantees and is the root cause that makes finding #1 harder to notice.

Suggested fix:
```python
class SubRequestedCheck(YotiSerializable, metaclass=ABCMeta):
```
(Note: this is the same pre-existing pattern in `RequestedCheck` and `YotiSerializable`; all three should be fixed together.)

**3. `.github/workflows/tests.yaml:69` — `pip install -e .[dev]` is unquoted, inconsistent with `sonar.yaml`**

`sonar.yaml` was updated to quote `".[dev]"` but the equivalent line in `tests.yaml` (unit-test job) was not. In shells where `nullglob` is set or where dot-files like `.d` / `.e` / `.v` exist, the glob `.[dev]` can expand unexpectedly and break the install step silently.

Suggested fix:
```yaml
- run: pip install -e ".[dev]"  # NOSONAR
```

---

### Nit

**4. `yoti_python_sdk/doc_scan/session/create/check/watchlist_screen.py:17–18` — Docstring `:type` field names are copy-paste errors**

`:type type: str` should be `:type manual_check: str`, and `:type max_retries: list` should be `:type categories: list`. These were copied from `liveness.py` and not updated.

**5. `yoti_python_sdk/doc_scan/session/create/subcheck/issuing_authority_sub_check.py:35–36` — Use `isinstance` instead of `issubclass(type(…), …)`**

```python
# current
if not issubclass(type(filter), DocumentFilter):
    raise ValueError('invalid filter')
```
`issubclass(type(x), T)` is equivalent to `isinstance(x, T)` for most cases but is less Pythonic, incompatible with virtual subclasses (`abc.register`), and is slightly harder to read. The parameter name `filter` also shadows the Python built-in.

Suggested fix:
```python
def with_filter(self, document_filter):
    if not isinstance(document_filter, DocumentFilter):
        raise ValueError('filter must be a DocumentFilter instance')
    self._filter = document_filter
    return self
```

**6. `yoti_python_sdk/doc_scan/session/retrieve/image_response.py:20–21` and `static_liveness_resource_response.py:24` — `"key" in data.keys()` should be `"key" in data`**

Both new files use `"media" in data.keys()` / `"image" in data.keys()`. The `.keys()` call is redundant — dict membership testing with `in` already checks keys. The adjacent `page_response.py` uses the same pattern, so this is a pre-existing style issue replicated into the new files.

PR #185, opened 2020-04-23
SDK-1458: Multiple documents and filters

MrBurtyyy:
## Added

* `yoti_python_sdk.doc_scan`
  * `session.create.filter` - Contains builders and objects used for specifying required documents during session creation
  * `DocScanClient`
    * `get_supported_documents()` - new method for retrieving all currently supported documents


yoti_python_sdk/doc_scan/session/create/filter/document_restrictions_filter.py
```diff
@@ -0,0 +1,154 @@
...
+        return self.__country_codes
+
+    @property
+    def document_types(self):
+        return self.__document_types
+
+    def to_json(self):
+        return {
+            "country_codes": self.country_codes,
+            "document_types": self.document_types,
```
davidgrayston, line 26:
```suggestion
            "country_codes": self.__country_codes,
            "document_types": self.__document_types,
```

Could we drop the getters and do this instead? It seems like these getters are only used for unit testing purposes, but would never be called directly otherwise _(this is to reduce the code we need to test/maintain)_

Also happy if it's kept as it is, as we have done similar in other SDKs and it doesn't hurt keeping the getters

yoti_python_sdk/doc_scan/session/create/filter/orthogonal_restrictions_filter.py
```diff
@@ -0,0 +1,178 @@
...
+                  .with_whitelisted_document_types(["PASSPORT"])
+                  .build())
+
+    """
+
+    def __init__(self):
+        self.__country_restriction = None
+        self.__type_restriction = None
+
+    def with_whitelisted_country_codes(self, country_codes):
```
davidgrayston, line 117:
```suggestion
    def with_whitelisted_countries(self, country_codes):
```

Should we use the same naming as Java? (I think both are fine and close enough)

> Same would apply elsewhere in this PR

yoti_python_sdk/doc_scan/session/create/filter/document_restrictions_filter.py
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+
+class DocumentRestrictionBuilder(object):
+    def __init__(self):
+        self.__country_codes = None
+        self.__document_types = None
+
+    def with_country_codes(self, country_codes):
```
davidgrayston, line 35:
```suggestion
    def with_countries(self, country_codes):
```

Perhaps this should match naming in Java? (I think it's close enough to leave it as is)

yoti_python_sdk/doc_scan/session/create/filter/document_restrictions_filter.py
davidgrayston: As discussed, we can leave this as is

yoti_python_sdk/doc_scan/session/create/filter/orthogonal_restrictions_filter.py
```diff
@@ -0,0 +1,178 @@
...
+
+        :return: the country restriction
+        :rtype: CountryRestriction
+        """
+        return self.__country_restriction
+
+    @property
+    def type_restriction(self):
+        """
+        Returns the document type restriction for the orthogonal filter]
```
davidgrayston, line 86:
```suggestion
        Returns the document type restriction for the orthogonal filter
```

yoti_python_sdk/doc_scan/session/create/filter/orthogonal_restrictions_filter.py
davidgrayston: As discussed, we're keeping them as they are

yoti_python_sdk/doc_scan/session/create/filter/document_restrictions_filter.py
davidgrayston: As discussed, we're keeping them as they are

yoti_python_sdk/doc_scan/session/create/filter/orthogonal_restrictions_filter.py
MrBurtyyy: Fixed

PR #241, opened 2020-09-04
SDK-1299: Update Doc Scan docs

echarrod:
As I was going through Go Doc Scan I noticed a few things that I thought should be changed


davidgrayston (approved):

echarrod:
@davidgrayston turning Sonar off and on again, and re-queuing the PR build caused the Sonar to be reported correctly. Did the same on PHP a few days ago and it didn't work, so must be a problem with the Sonar check's comms 🤔

PR #244, opened 2020-09-15
SDK-1718: Remove null values from JSON payload

echarrod:
Even if the objects don't have any values which currently serialize to null, I thought it would be best to specify that we don't want this behaviour, and then if we add any in the future, they will be excluded if null.

The alternative to this would be to ignore these values by default, and then we can override in any methods where we need to.

### DIFF
--- a/yoti_python_sdk/doc_scan/session/create/check/face_match.py
+++ b/yoti_python_sdk/doc_scan/session/create/check/face_match.py
@@ -32,6 +32,9 @@ class RequestedFaceMatchCheckConfig(YotiSerializable):
     def to_json(self):
         return {"manual_check": self.manual_check}
 
+    def include_null_values(self):
+        return False
+
 
 class RequestedFaceMatchCheck(RequestedCheck):
     """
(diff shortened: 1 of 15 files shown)


yoti_python_sdk/doc_scan/session/create/check/face_match.py
```diff
@@ -32,6 +32,9 @@ def manual_check(self):
     def to_json(self):
         return {"manual_check": self.manual_check}
 
+    def include_null_values(self):
+        return False
+
```
davidgrayston, line 37:
```suggestion
    def to_json(self):
        return remove_null_values({
            "manual_check": self.manual_check,
        })

```

davidgrayston (approved):
An alternative would be to call `remove_null_values` in each `to_json` method, which would prevent exposing a new public method on each object. Not sure if that would be better or worse?

yoti_python_sdk/doc_scan/session/create/check/face_match.py
echarrod: I was thinking we're a bit clearer with our intentions if we leave them as separate, but I think my preference to not have clients exposed to the include_null_values method overrides that, so I've changed in ece2cad

echarrod:
@davidgrayston only thing I wasn't sure about in ece2cad was having:
```python
    def to_json(self):
        return remove_null_values({})
```
for RequestedDocumentAuthenticityCheckConfig. It isn't doing anything since there are not configuration options at the moment, but I thought it was best to leave in, in case we add any further objects to this config

davidgrayston (approved):

PR #246, opened 2020-09-17
SDK-1582: Include response error in exception message

davidgrayston:
### Added
- Formatted response error is now included in `DocScanException` message

### DIFF
--- a/yoti_python_sdk/doc_scan/exception/doc_scan_exception.py
+++ b/yoti_python_sdk/doc_scan/exception/doc_scan_exception.py
@@ -1,3 +1,6 @@
+import json
+
+
 class DocScanException(Exception):
     """
     Exception thrown by the Yoti Doc Scan client
@@ -13,7 +16,11 @@ class DocScanException(Exception):
         """
         Exception.__init__(self)
 
-        self.__message = message
+        response_message = self.__format_response_message(response)
+
+        self.__message = message + (
+            " - " + response_message if response_message else ""
+        )
         self.__response = response
 
     @property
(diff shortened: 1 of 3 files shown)


yoti_python_sdk/tests/mocks.py
```diff
@@ -11,7 +12,9 @@ def __init__(self, status_code, text, headers=None, content=None):
         if headers is None:
             headers = dict()
 
-        super(MockResponse, self).__init__(status_code, text, headers, content)
+        super(MockResponse, self).__init__(
+            status_code, text, CaseInsensitiveDict(headers), content
```
davidgrayston, line 16: I've used `CaseInsensitiveDict` here to mock the behaviour of _Requests_ case insensitive headers. It doesn't affect existing tests, but I thought it would be best for the mock headers to behave in the same way.

yoti_python_sdk/doc_scan/exception/doc_scan_exception.py
```diff
@@ -56,5 +63,32 @@ def content(self):
         """
         return self.__response.content
 
+    def __format_response_message(self, response):
```
echarrod, line 66: Do you think we should have `if response.headers.get("Content-Type") == "application/json":` outside of this function? And then we can call this `__format_json_response_message` to better reflect what it's doing.

davidgrayston: I've split this method in e64441c - I think it's a little bit clearer now

echarrod (approved):

PR #248, opened 2020-09-21
IN-3118: python Example QR

nikhilPank:


echarrod (approved):

davidgrayston (changes requested):
Do we need to make this change in the dynamic share example too?

https://github.com/getyoti/yoti-python-sdk/blob/development/examples/yoti_example_django/yoti_example/templates/dynamic-share.html#L73

https://github.com/getyoti/yoti-python-sdk/blob/development/examples/yoti_example_flask/templates/dynamic-share.html#L73

davidgrayston (approved):

PR #250, opened 2020-10-02
SDK-1732: Allow manual check to be configured on authenticity checks

davidgrayston:
### Added
- Allow the configuration of the `manual_check` value for requested Document Authenticity checks

### Fixed
- `RequestedDocumentAuthenticityCheckBuilder.build()` is no longer static. Python 2 implementations must change this to `RequestedDocumentAuthenticityCheckBuilder().build()`

### DIFF
--- a/yoti_python_sdk/doc_scan/session/create/check/document_authenticity.py
+++ b/yoti_python_sdk/doc_scan/session/create/check/document_authenticity.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from yoti_python_sdk.doc_scan.constants import ID_DOCUMENT_AUTHENTICITY
+from yoti_python_sdk.doc_scan import constants
 from yoti_python_sdk.utils import YotiSerializable, remove_null_values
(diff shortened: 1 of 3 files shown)


yoti_python_sdk/doc_scan/session/create/check/document_authenticity.py
```diff
@@ -41,7 +59,44 @@ class RequestedDocumentAuthenticityCheckBuilder(object):
...
+        :rtype: RequestedDocumentAuthenticityCheckBuilder
+        """
+        self.__manual_check = constants.NEVER
+        return self
+
+    def build(self=None):
+        if self is None:
+            manual_check = None
+        else:
+            manual_check = self.__manual_check
```
davidgrayston, line 99: Previously `@staticmethod` which we need to support for backwards compatibility (we can remove this in the major release)

davidgrayston: This cannot work with Python 2, so we need to deprecate this builder and introduce a new one

echarrod (approved):
👍

PR #310, opened 2021-07-09
adding myself to dependabot

MichalMilewiczYoti:
adding myself to dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,4 @@ updates:
   - gautam-yoti
   assignees:
   - echarrod
+  - michalmilewiczyoti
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -11,3 +11,4 @@ updates:
   - gautam-yoti
   assignees:
   - echarrod
```
echarrod, line 13: We try to usually have one "assignee", so that one person is responsible for merging:
```suggestion
```

.github/dependabot.yml
```diff
@@ -11,3 +11,4 @@ updates:
   - gautam-yoti
```
echarrod, line 11: And we can add you as a reviewer too
```suggestion
  - gautam-yoti
  - michalmilewiczyoti
```

echarrod (approved):

PR #321, opened 2021-08-31
SDK-1901: Adding WATCHLIST_SCREENING check

MichalMilewiczYoti:
SDK-1901 adding WATCHLIST_SCREENING check


ShamariYoti (approved):
Looks good, If you could add a : after the ticket in the commit and PR title so we can link the commints and PR to JIRA.

yoti_python_sdk/tests/test_anchor.py
```diff
@@ -121,7 +121,10 @@ def test_processing_unknown_anchor_data():
     expected_timestamp = datetime(2019, 3, 5, 10, 45, 11, 840037)
     actual_timestamp = anchors[0].signed_timestamp
 
-    assert expected_timestamp == actual_timestamp
+    try:
+        assert expected_timestamp == actual_timestamp
+    except AssertionError:
+        assert expected_timestamp == actual_timestamp - timedelta(hours=1)
```
echarrod, line 127: We need to find a way of doing this independent of timezone

echarrod (changes requested):

yoti_python_sdk/tests/test_anchor.py
MichalMilewiczYoti: yes it will be easlier, let's drop support for python2.7

echarrod: It has been dropped in development, if you do a rebase or a merge you'll get the updated Github actions file which doesn't have 2.7 (looks like there's some conflicts, so I think you'll need to do this anyway)

echarrod: Looks good now :+1:

echarrod (approved):
Nice work Michal :ok_hand:

PR #324, opened 2021-09-23
SDK-1838 auth_type for doc_scan session

MichalMilewiczYoti:
adding auth_type for doc_scan session related with notification


yoti_python_sdk/doc_scan/session/create/notification_config.py
```diff
@@ -62,9 +63,21 @@ def topics(self):
         """
         return self.__topics
 
+    @property
+    def auth_type(self):
+        """
+        The authentication type that notification will used to
```
echarrod, line 69:
```suggestion
        The authentication type that notification will use to
```

MichalMilewiczYoti: done

yoti_python_sdk/doc_scan/session/create/notification_config.py
```diff
@@ -62,9 +63,21 @@ def topics(self):
         """
         return self.__topics
 
+    @property
+    def auth_type(self):
+        """
+        The authentication type that notification will use to
```
ShamariYoti, line 69:
```suggestion
        The authentication type that the notification will use to
```

ShamariYoti (commented):
Just one small suggestion for me

yoti_python_sdk/doc_scan/session/create/notification_config.py
```diff
@@ -16,7 +16,7 @@ class NotificationConfig(YotiSerializable):
     to poll for the state of the Session.
     """
 
-    def __init__(self, auth_token, endpoint, topics=None):
+    def __init__(self, auth_token, endpoint, auth_type, topics=None):
```
echarrod, line 19: I know we have builders and so we're not recommending clients initialise with this, but at all times when adding to functions we try to:
1) put new parameters at the end of the parameter list. If it's as it is currently, then any existing calls to this function which are passing `topics` will now pass this as the `auth_type`
2) We initialise as `None`, so that clients aren't _required_ to set this parameter

yoti_python_sdk/doc_scan/session/create/notification_config.py
```diff
@@ -81,6 +94,7 @@ def __init__(self):
         self.__auth_token = None
         self.__endpoint = None
         self.__topics = []
+        self.__auth_type = "BASIC"
```
echarrod, line 97: Why are we hardcoding to `BASIC` here? It should be `None`, right?

yoti_python_sdk/tests/doc_scan/session/create/test_notification_config.py
```diff
@@ -26,6 +27,7 @@ def test_should_build_correctly(self):
         assert result.auth_token is self.SOME_AUTH_TOKEN
         assert result.endpoint is self.SOME_ENDPOINT
         assert self.SOME_TOPIC in result.topics
+        assert result.auth_type is "BASIC"
```
echarrod, line 30: I think we want this to be `None` if it is not set, so we can distinguish between unset, and intentionally set to `BASIC`

echarrod (changes requested):

ShamariYoti (approved):
LGTM, would be only to address Ed's comments

echarrod (approved):
Looks good now Michal, I think we need a rebase or merge from master to sort the conflicts though :+1:

echarrod:
Oh actually, tests are failing. Once those are fixed I'm happy to merge this

echarrod:
@MichalMilewiczYoti I've merged the  branch in which fixed the Github actions, and rebased onto development to avoid the conflicts. Please can you check the new changes all look OK after the rebase, before merging :+1:

PR #343, opened 2021-12-06
[SDK-1972] - Add support for Face Capture resource

MichalMilewiczYoti:
Add support for Face Capture resource


yoti_python_sdk/doc_scan/session/retrieve/face_capture_response.py
```diff
@@ -0,0 +1,28 @@
...
+    def source(self):
+        return self._data.get("source")
+
+    @property
+    def image(self):
+        return MediaResponse(data=self._data.get("image") or {})
+
+    @property
+    def tasks(self):
+        return self._data.get("tasks") or []
```
ShamariYoti, line 28: Does this line do anything, correct me if ii'm wrong but won't this always return `self._data.get("tasks")` ?

```suggestion
        return self._data.get("tasks") if len(self._data.get("tasks")) > 0 else []
```

ShamariYoti (commented):
just one question

yoti_python_sdk/doc_scan/session/retrieve/face_capture_response.py
MichalMilewiczYoti: in python return self._data.get("tasks") or [] will work same like Your solution:

self._data.get("tasks") -> if its [] or None it will use shortcut value -> []
return self._data.get("tasks") if len(self._data.get("tasks")) > 0 else []

in python one liners are consinder as anti pattern. In 90 % cases You can use shortcuts that are more readable and user friendly.

Summary - current state of the code works the same like Your proposition

echarrod (changes requested):
We'll need to add some more things for this I believe, please see the [.NET implementation](https://github.com/getyoti/yoti-dotnet-sdk/pull/332/files) as an example

yoti_python_sdk/doc_scan/session/retrieve/face_capture_response.py
ShamariYoti: Cool thanks for clarifying.

ShamariYoti (approved):

PR #365, opened 2022-06-13
Celery >=5.2.1 compatibility to unblock CVE patch

ZephSibley:
yoti-python-sdk has a pytz dependency that prevents upgrading Celery past 5.2 - this blocks projects that use both libraries from patching the following high severity CVE: https://github.com/advisories/GHSA-q4xr-rc97-m4xx

Please consider this PR as soon as you are able.

Thanks!


echarrod (approved):
Thanks for the contribution @ZephSibley. I've updated the base branch to be 2.14.1, and we will release this in the 2.14.1 release which will go out next week.

CI checks are failing on this one also, but we will address in a separate PR.

PR #453, opened 2026-04-07
SDK-2743-python-expose-idv-breakdown-process-property

mehmet-yoti:
## Summary
  - Add `process` property to `BreakdownResponse` to expose the breakdown process type
  (AUTOMATED / EXPERT_REVIEW)
  - Field is a nullable string, defaults to None when absent


nikhilPank (approved):
LGTM

PR #460, opened 2026-05-11
SDK-2781: Python - Add support for new capture_type property on Static Liveness resources - python

mehmet-yoti:
## Summary
Adds support for the new `capture_type` property on Static Liveness resources returned by the Doc Scan API. The `StaticLivenessResourceResponse` class now exposes a `capture_type` attribute (e.g. `"PHOTOGRAPH"`) parsed from the API response, defaulting to `None` when the field is absent.

## Changes
- `yoti_python_sdk/doc_scan/session/retrieve/static_liveness_resource_response.py`
  - Parse `capture_type` from the response payload in `__init__` using `data.get("capture_type", None)`.
  - Add a new `capture_type` `@property` exposing the value (returns `str` or `None`).
- `yoti_python_sdk/tests/doc_scan/session/retrieve/test_static_liveness_resource.py`
  - Updated existing happy-path test to assert `capture_type == "PHOTOGRAPH"`.
  - Asserted `capture_type is None` in the existing missing-image test.
  - Added `test_should_handle_missing_capture_type` and `test_should_handle_none_data` to cover absent/empty payloads.
- `yoti_python_sdk/tests/doc_scan/session/retrieve/test_resource_container.py`
  - Added `test_should_expose_capture_type_on_static_liveness_resource` verifying that `capture_type` is wired through `ResourceContainer.static_liveness_resources` and is not present on non-static (Zoom) liveness resources.

## QA Test Steps
1. **Setup**
   - Check out the branch `websdk-auto/SDK-2781-python-python---add-support-for-new-capture-type-property-on-static-liveness-resources`.
   - Install dev dependencies: `pip install -e .` and `pip install -r requirements.txt` (or your usual install command).
2. **Run the new and updated unit tests**
   - `pytest yoti_python_sdk/tests/doc_scan/session/retrieve/test_static_liveness_resource.py -v`
   - `pytest yoti_python_sdk/tests/doc_scan/session/retrieve/test_resource_container.py -v`
   - Expect all tests to pass, including:
     - `test_should_parse_static_liveness_resource_response` (asserts `capture_type == "PHOTOGRAPH"`).
     - `test_should_handle_missing_capture_type` and `test_should_handle_none_data`.
     - `test_should_expose_capture_type_on_static_liveness_resource`.
3. **Run the full doc_scan test suite to check for regressions**
   - `pytest yoti_python_sdk/tests/doc_scan/ -v`
   - Confirm no other tests fail.
4. **Happy path — capture_type present**
   - In a Python REPL:
     ```python
     from yoti_python_sdk.doc_scan.session.retrieve.static_liveness_resource_response import StaticLivenessResourceResponse
     r = StaticLivenessResourceResponse({"id": "x", "liveness_type": "STATIC", "capture_type": "PHOTOGRAPH"})
     assert r.capture_type == "PHOTOGRAPH"
     ```
5. **Edge case — capture_type missing**
   - `StaticLivenessResourceResponse({"id": "x", "liveness_type": "STATIC"}).capture_type` must be `None`.
6. **Edge case — None payload**
   - `StaticLivenessResourceResponse(None).capture_type` must be `None` (and `.image` must be `None`).
7. **Integration via ResourceContainer**
   - Build a `ResourceContainer` with a `liveness_capture` array containing one `STATIC` entry with `capture_type` and one `ZOOM` entry; confirm `static_liveness_resources[0].capture_type` returns the expected value and the Zoom resource has no `capture_type` attribute.
8. **Regression check**
   - Verify other `StaticLivenessResourceResponse` fields (`id`, `liveness_type`, `image`, `image.media.id`) still parse correctly when `capture_type` is both present and absent.

## Notes
- `capture_type` is exposed as a plain string (e.g. `"PHOTOGRAPH"`) rather than an enum, mirroring how other backend-driven string fields (`liveness_type`, source `type`, etc.) are handled in this SDK. If a canonical set of values is later defined, an enum could be added in a follow-up without breaking the public API.
- The field defaults to `None` when missing, preserving backwards compatibility with older API responses that do not include `capture_type`.
- No public API was removed and no existing behaviour changed — this is a purely additive change.
- Only the Python `StaticLivenessResourceResponse` is updated; other liveness resource types (e.g. Zoom) are intentionally untouched per the ticket scope.

---

*Related Jira:* SDK-2781
*Auto-generated by n8n + Claude CLI*


mehmet-yoti:
## 🤖 Claude Code Review

## Code Review Findings

### Critical
None

### Major
None

### Minor
- `yoti_python_sdk/tests/doc_scan/session/retrieve/test_resource_container.py:85` — The assertion `assert not hasattr(zoom_resources[0], "capture_type")` is fragile and slightly misleading. `ZoomLivenessResourceResponse` inherits from `LivenessResourceResponse`, not from `StaticLivenessResourceResponse`, so `capture_type` will of course not be present — but this assertion would silently pass even if the property were ever added to the base class. It's not really testing the feature; the meaningful coverage already lives in `test_static_liveness_resource.py`. Either drop this assertion or replace it with a more direct check that the zoom resource is not a `StaticLivenessResourceResponse`, e.g.:
  ```python
  assert not isinstance(zoom_resources[0], StaticLivenessResourceResponse)
  ```

- `yoti_python_sdk/doc_scan/session/retrieve/static_liveness_resource_response.py:26` — The rest of this file uses the `"key" in data.keys()` idiom for the `image` field, while the new line uses `data.get("capture_type", None)`. Both work, but the sibling `LivenessResourceResponse.__init__` (liveness_resource_response.py:20) also uses `data.get(..., None)`, so the new line is consistent with the parent class. Consider unifying within this file for readability — either leave the new line as is (matches parent) or align the `image` line on the same idiom:
  ```python
  self.__image = ImageResponse(data["image"]) if "image" in data else None
  self.__capture_type = data.get("capture_type")
  ```
  No behavior change either way.

### Nit
- `yoti_python_sdk/doc_scan/session/retrieve/static_liveness_resource_response.py:26` — `data.get("capture_type", None)` is equivalent to `data.get("capture_type")`; the explicit `None` default is redundant. Minor style.
- `yoti_python_sdk/doc_scan/session/retrieve/static_liveness_resource_response.py:40-45` — The docstring summary is fine, but for parity with `image` you could add a brief sentence describing the expected values (e.g., `"PHOTOGRAPH"`), since `capture_type` is documented as `str or None` without enumerating known values. Optional.
- `yoti_python_sdk/tests/doc_scan/session/retrieve/test_static_liveness_resource.py:71-75` — `test_should_handle_none_data` is a nice addition; consider also asserting `result.id is None` and `result.liveness_type is None` for completeness, matching the style of `test_should_handle_missing_image`.

J-Stead (approved):
LGTM

saurabh-yoti (approved):

nikhilPank (approved):

